PR #10, opened 2021-04-28
9 integrated eslint, upgraded react-script, aligned code with eslint …

chrisahardie:
…config


chrisahardie (commented):
Here is a branch with eslint integrated (along with prettier) with formatting applied to the existing codebase. I included a number of rules I have found helpful, but everything can certainly be changed in the config.

@ericherscovich I suggest pulling the branch down, installing the new packages, and make sure things work/are suitable for you in your IDE of choice. Most IDEs have Eslint plugins. I have found eslint to be beneficial esp when you have many devs touching a codebase, but i'll leave it to you to decide whether there is value in this PR. If so, let me know and I'll update the README with some guidance.

package.json
```diff
@@ -4,22 +4,22 @@
...
   "dependencies": {
     "@shopify/polaris": "^4.19.0",
+    "@shopify/polaris-icons": "^4.5.0",
     "@testing-library/jest-dom": "^4.2.4",
     "@testing-library/react": "^9.3.2",
     "@testing-library/user-event": "^7.1.2",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
-    "react-scripts": "3.4.1",
+    "react-scripts": "4.0.3",
```
chrisahardie, line 13: It appears the codebase was possible produced from an older js-oriented codebase? I upgraded to the latest version of CRA. `testing-library` should be updated as well, but i didn't want to touch even more in this ticket.

src/App.tsx
```diff
@@ -1,31 +1,36 @@
 import React, { useCallback, useState } from "react";
 import "@shopify/polaris/styles.css";
 import { Frame, TopBar, Navigation } from "@shopify/polaris";
-import {
-  HomeMajorMonotone,
-} from "@shopify/polaris-icons";
-import Home from "./components/Home";
+import { HomeMajor } from "@shopify/polaris-icons";
```
chrisahardie, line 4: Compiler couldn't find `HomeMajorMonotone` in `@shopify/polaris-icons`, wouldn't compile ...

src/App.tsx
```diff
@@ -1,31 +1,36 @@
...
-function App() {
+export function App() {
   const [menuState, setMenuState] = useState("home");
   const setPageMarkup = useCallback((value) => setMenuState(value), []);
 
   const userMenuMarkup = (
-    <TopBar.UserMenu initials="EH" />
+    <TopBar.UserMenu
+      initials="EH"
+      actions={[]}
```
chrisahardie, line 14: Adding a number of mandatory but useless props, compiler won't compile without them

package.json
```diff
@@ -4,22 +4,22 @@
...
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test",
-    "eject": "react-scripts eject"
-  },
-  "eslintConfig": {
-    "extends": "react-app"
+    "eject": "react-scripts eject",
+    "lint": "eslint --max-warnings 0 --ext .tsx,.ts ./src",
```
chrisahardie, line 21: This is useful for devs, but can also be used in husky hooks and CI servers - I generally run a lint script, then a test script, then the build script before pushing code.

chrisahardie (commented):
Oh, there is also `yarn.lock` and `package-lock.json` - wasn't sure which package manager you are using, but one of these files should go.

Issue #11, opened 2021-04-28
Add husky hook


chrisahardie:
Add a hook to ensure code style and quality is maintained. I would suggest the hook runs:

* the linting script
* tests
* the build script

PR #17, opened 2021-04-30
Add pre-commit hooks and fix tests

johnhok:
Utilize [lint-staged](https://github.com/okonet/lint-staged) to add pre-commit hooks to auto format code with prettier to ensure code style consistency.

Personally I don't think the build and tests should run in the pre-commit hook. I've always been in favour of pre-commit hooks running fast. CI should be validating the build and tests. But open to adding tests to the pre-commit.

Closes: https://github.com/Vaccine-Hunters-Canada/VaxFinder-Frontend/issues/11


src/App.tsx
```diff
@@ -70,8 +80,10 @@ export function App() {
   };
 
   return (
-    <Frame topBar={topBarMarkup} navigation={navigationMarkup}>
-      {getCurrentMarkup()}
-    </Frame>
+    <AppProvider theme={theme} i18n={enTranslations}>
```
johnhok, line 75: The tests were failing without the `AppProvider` at the root level.

chardielxr:
> Utilize [lint-staged](https://github.com/okonet/lint-staged) to add pre-commit hooks to auto format code with prettier to ensure code style consistency.
> 
> Personally I don't think the build and tests should run in the pre-commit hook. I've always been in favour of pre-commit hooks running fast. CI should be validating the build and tests. But I can update the PR if we're in favour to adding tests as well in the the pre-commit.
> 
> Closes: #11

I prefer pre-push hooks over pre-commit as I also want my commits to be fast :) We don't have CI now, so nowhere else presently would ensure our tests pass and our code compiles.

mittonface:
> We don't have CI now, so nowhere else presently would ensure our tests pass and our code compiles.

This doesn't seem like a really big concern. https://github.com/Vaccine-Hunters-Canada/VaxFinder-Frontend/pull/18

package.json
```diff
@@ -51,6 +53,12 @@
...
     "eslint-plugin-react-hooks": "1.7.0",
     "eslint-plugin-testing-library": "3.3.1",
-    "prettier": "^2.2.1"
+    "husky": ">=6",
+    "lint-staged": ">=10",
+    "prettier": "^2.2.1",
+    "pretty-quick": "^3.1.0"
+  },
+  "lint-staged": {
+    "*.{tsx,ts}": "pretty-quick --staged"
```
chardielxr, line 62: This is just running prettier? Why not run the lint script? Prettier has been integrated via the eslint plugin...

chardielxr, line 62: Is this just running prettier? Prettier is integrated into eslint via a plugin. Can we run the lint script instead?

.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
```
chardielxr, line 1: I would move all these rules into the eslint config.... prettier is integrated into eslint.

johnhok:
> > Utilize [lint-staged](https://github.com/okonet/lint-staged) to add pre-commit hooks to auto format code with prettier to ensure code style consistency.
> > Personally I don't think the build and tests should run in the pre-commit hook. I've always been in favour of pre-commit hooks running fast. CI should be validating the build and tests. But I can update the PR if we're in favour to adding tests as well in the the pre-commit.
> > Closes: #11
> 
> I prefer pre-push hooks over pre-commit as I also want my commits to be fast :) We don't have CI now, so nowhere else presently would ensure our tests pass and our code compiles.

Are you thinking the push would fail and you'd go lint the files, create a commit and try to re-push? Or are you thinking of some other way to auto fix before push?

package.json
chardielxr: I don't see any mention on eslint in the pretty-quick docs - you sure it is linting?

johnhok: You're right, it's not linting. I'll change the hook to run the lint task.

.eslintrc.json
```diff
@@ -1,235 +1,219 @@
...
+  "rules": {
+    "no-use-before-define": "off",
+    "prettier/prettier": [
+      "error",
+      {
+        "semi": true,
+        "trailingComma": "all",
+        "singleQuote": false,
+        "printWidth": 80
+      }
```
johnhok, line 45: This file was auto formatted... this is the pertinent change.

.husky/pre-push
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npm run lint
```
chrisahardie, line 4: Wow, they have really changed Husky hooks!

johnhok: They have! Was quite confused at first.

.eslintrc.json
chrisahardie: noice - one config and script to rule them all :)

patcon:
Another option: https://github.com/reviewdog/action-suggester

screenshot of reviewdog suggester action

Basically, the `--fix` command just gets run in a CI step, creating a dirty git workspace, and the action above creates github review comments (with code suggestions) for each hunk of the diff. Accepting them can just be done in bulk via github UI, like any code suggestions. (tl;dr Removes the back-and-forth of "hey new contributor, can you fix this and we'll merge?")

I can spin this out into an other PR if there are +1's but it's outta scope here 😄

patcon (approved):

chrisahardie:
> fwiw: I _love_ automatic fixing. For me, part of the entire purpose of husky and auto fixing in a hook is not having to troubleshoot other devs as much!

Devs should be responsible for the changes they are introducing. Any issues raised via a husky hook or CI script failure should be attended to by the dev that caused the failure. There should be no "troubleshooting" other devs, I usually don't look at a PR until i know quality checks have passed.

chrisahardie (approved):

chrisahardie:
> re: where/how linting happens. Another option: GitHub PRs via CI https://github.com/reviewdog/action-suggester
> 
> (I can happily spin this out into an other PR if there are +1's and it's outta scope here 😄 )
> 
> screenshot of reviewdog suggester action
> 
> Basically, an arbitrary linting `--fix` command just gets run in a CI step, creating a dirty git workspace, and the action above creates github review comments (with code suggestions) for each hunk of the diff. Accepting them can just be done in bulk via github UI, like any code suggestions. (tl;dr Removes the back-and-forth of "hey new contributor, can you fix this and we'll merge?")

Neat - I'll let others weigh in on this, I'm still a Github noob.

johnhok:
> re: where/how linting happens. Another option: GitHub PRs via CI https://github.com/reviewdog/action-suggester
> 
> (I can happily spin this out into an other PR if there are +1's and it's outta scope here 😄 )
> 
> screenshot of reviewdog suggester action
> 
> Basically, an arbitrary linting `--fix` command just gets run in a CI step, creating a dirty git workspace, and the action above creates github review comments (with code suggestions) for each hunk of the diff. Accepting them can just be done in bulk via github UI, like any code suggestions. (tl;dr Removes the back-and-forth of "hey new contributor, can you fix this and we'll merge?")

The pre-push hook will prevent code that is failing lint to even be pushed in the first place? This should prevent back and forth about linting/styling concerns.

patcon:
> The pre-push hook will prevent code that is failing lint to even be pushed in the first place? This should prevent back and forth about linting/styling concerns.

:+1: seems comparable

I generally have loose open source bias toward [automated] things happening in collectivized and shared/observable spaces (PRs, chat, CI), rather than spaces we wrangle with alone (localhost). E.g. chatbots vs CLI commands

PR #18, opened 2021-04-30
Add build test lint action

mittonface:
Probably worth noting that this is failing at the moment. But this is an instance of the workflow doing its job.

https://github.com/mittonface/VaxFinder-Frontend/runs/2476308686?check_suite_focus=true


.github/workflows/build-test-lint.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Install
+        run: yarn install
+
+      - name: Build
+        run: yarn build
+
+      - name: Test
+        run: yarn test --watchAll=false 
+
+      - name: Lint
```
chardielxr, line 25: Probably doesn't matter for the size of the codebase, but I generally move lint to the top. Linting fails relatively fast, I have test suites that can take over 15 minutes to run :)

chardielxr, line 25: Probably doesn't matter for the size of the codebase, but I generally move lint to the top. Linting fails relatively fast, I have test suites that can take over 15 minutes to run :)

mittonface: This is low-key genius! I've never thought of this.

.github/workflows/build-test-lint.yml
```diff
@@ -0,0 +1,28 @@
+
```
chrisahardie, line 1: I'm not familiar with actions, but should we be setting the CI env var to true? It's a matter of course for all my CI scripts.

patcon: Good call, but I believe GH Actions already sets this as default, so no need to specify: https://docs.github.com/en/actions/reference/environment-variables

Good eyes though 👀

chrisahardie:
@mittonface Wanna pull from `main` and see if your shiny new action will give us a greenlight?

chrisahardie (approved):

chrisahardie:
Thanks @mittonface , I believe this is working a treat! https://github.com/mittonface/VaxFinder-Frontend/actions/runs/800022702

patcon:
Yes, I concur: totally a treat, indeed 🍬

PR #24, opened 2021-05-01
Prompts for postal code

picuron:


src/components/Home/Home.tsx
```diff
@@ -27,18 +41,67 @@ export function Home() {
...
+  const handleSubmit = () => {
+    const postalCodeRegex = /[a-zA-Z][0-9][a-zA-Z](-| |)[0-9][a-zA-Z][0-9]/;
+    if (postalCodeRegex.test(postalCode)) {
+      setRenderCards(true);
+      setShowPostalPrompt(false);
+    } else {
+      setShowInvalidPostal(true);
+    }
+  };
+  const afterPostalEntryMarkup = renderCards ? (
```
luceliu, line 54: `renderCards && (...)` i think would be more concise tho this totally works

src/components/Home/Home.tsx
```diff
@@ -27,18 +41,67 @@ export function Home() {
...
+    <>
+      <TextStyle variation="subdued">
+        Here&apos;s what&apos;s happening with vaccinations in Ottawa
+      </TextStyle>
+      <br />
```
luceliu, line 59: @moni-li curious abt your thoughts re: <br/> for styling w.r.t. accessibility. my understanding is that <br>s are used for semantic breaks in text

src/components/Home/Home.tsx
chrisahardie: > `renderCards && (...)` i think would be more concise tho this totally works

Yes, you are correct, and the official docs even mention using the logical AND operator for conditional rendering, but there is a case to be made for using ternaries: https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

For consistency's sake, I always use ternaries.

src/components/Home/Home.tsx
```diff
@@ -1,10 +1,24 @@
...
+  Button,
+  Card,
+} from "@shopify/polaris";
 import React, { useState } from "react";
 import { PharmacyList } from "../PharmacyList";
 import { TwitterTimelineEmbed } from "react-twitter-embed";
 
 export function Home() {
   const [showBanner, setShowBanner] = useState(true);
+  const [showPostalPrompt, setShowPostalPrompt] = useState(true);
```
chrisahardie, line 18: In any programming language, it is typical to prefix booleans variables. The most common prefix `is`, but other conventions include with `are`, `has`, `should`, `can`

I have found it to be an extremely helpful cue, especially in JSX.

src/components/Home/Home.tsx
chrisahardie: > @moni-li curious abt your thoughts re: `<br/>` for styling w.r.t. accessibility. my understanding is that `<br/>`s are used for semantic breaks in text

The ideal is to not use `<br />` for stylistic purposes - e.g. to add whitespace

src/components/Home/Home.tsx
```diff
@@ -27,18 +41,67 @@ export function Home() {
...
+    <>
+      <TextStyle variation="subdued">
+        Here&apos;s what&apos;s happening with vaccinations in Ottawa
+      </TextStyle>
+      <br />
+      <br />
+      {dismissableBannerMarkup}
+      <PharmacyList />
+    </>
+  ) : null;
```
chrisahardie, line 64: Don't bother doing anything about this, but I never use `null` in my code, ever. JS is weird in that it has two competing bottom types. I always use `undefined`, as that is the bottom type the language itself uses. https://www.youtube.com/watch?v=PSGEjv3Tqo0&t=561s

src/components/Home/Home.tsx
```diff
@@ -27,18 +41,67 @@ export function Home() {
...
+
+  const postalLabelMarkup = showInvalidPostal
```
chrisahardie, line 66: On error, we appear to not be calling out the issue stylistically or via a semantic accessible markup. We should set the input's `aria-invalid` attribute  to `true`. Also add `aria-describedby` to the input on error, and reference the error element:

```
<TextField name="postal"
		id="postal"
                aria-invalid
		aria-describedby="postalError">
		<span id="postalError">You have entered an invalid postal code. Please enter a valid one in this format: K2T 0E5</span>

File: src/components/Home/Home.tsx [hunk 6]
```diff
@@ -1,10 +1,24 @@
...
+} from "@shopify/polaris";
 import React, { useState } from "react";
 import { PharmacyList } from "../PharmacyList";
 import { TwitterTimelineEmbed } from "react-twitter-embed";
 
 export function Home() {
   const [showBanner, setShowBanner] = useState(true);
+  const [showPostalPrompt, setShowPostalPrompt] = useState(true);
+  const [showInvalidPostal, setShowInvalidPostal] = useState(false);
+  const [renderCards, setRenderCards] = useState(false);
```
chrisahardie, line 20: Would `shouldRenderPharmacyList` would be more precise? "cards" has no semantic meaning.

chrisahardie (commented):
There is logic here we could exercise with tests. Have we landed on a testing strategy yet? I'm a big fan of `testing-library`, and could set that up if desired.

src/components/Home/Home.tsx
picuron: Removed All `<br />` tags.

src/components/Home/Home.tsx
picuron: Changed to "shouldShow..."

src/components/Home/Home.tsx
picuron: Is the way I have it set up with postalLabelMarkup not valid? If a postal is invalid showInvalidPostal is true and shows the message in the title of the textbox that they entered an invalid postal code.

chrisahardie: The label is no longer describing its related control, which might be confusing. I believe you can add the error message INSIDE the label as additional information, but the label should still be responsible for indicating what data the control is responsible for.

Also, the error isn't styled to indicate that the control is in an error state. Error stylings would include things like red text, a red outline, a big ol' X... Those visual cues are helpful to everyone, but especially those with visual and cognitive impairments.

src/components/Home/Home.tsx
```diff
@@ -23,28 +38,68 @@ export function Home() {
...
+                value={postalCode}
+                onChange={(postal) => setPostalCode(postal)}
+                label="Please enter your postal code (Example: K2T 0E5):"
+                helpText={
+                  <span>
+                    We’ll use this postal code to find the closest available
+                    vaccines.
+                  </span>
+                }
+                error={invalidPostalCodeMessage}
```
chrisahardie, line 80: @picuron By virtue of adding a message to the error prop, `TextField` will take care of showing an inline error and setting all the aria properties.

chrisahardie (dismissed):

src/components/Home/Home.tsx
```diff
@@ -23,28 +38,68 @@ export function Home() {
...
+
+  const handleSubmit = () => {
+    const postalCodeRegex = /[a-zA-Z][0-9][a-zA-Z](-| |)[0-9][a-zA-Z][0-9]/;
+    if (postalCodeRegex.test(postalCode)) {
+      setShouldRenderPharmacyList(true);
+      setShouldShowPostalPrompt(false);
+    } else {
+      setShouldShowInvalidPostal(true);
+    }
+  };
```
johnhok, line 52: Instead of setting a local state, do you think it's worthwhile to create a dedicated route for this? The problem is if the user refreshes, they'll get sent back to this postal code prompt.

johnhok:
> There is logic here we could exercise with tests. Have we landed on a testing strategy yet? I'm a big fan of `testing-library`, and could set that up if desired.

👍  for testing library

src/components/Home/Home.tsx
chrisahardie: A route with the postal code as a route param would be nice so we could deep link a URL straight to the results screen.

There was also talk about using localStorage to store the postal code so users wouldn't have to re-enter it, so if we implement both of these flows, we would have they don't cross swords.

chrisahardie (approved):

johnhok (approved):

Issue #21, opened 2021-05-01
Set up Typescript objects based on Swagger schema


picuron:
Look into using https://github.com/acacode/swagger-typescript-api for this. 

Note that the Swagger docs may be different tomorrow (May 1).

johnhok:
Closed by https://github.com/Vaccine-Hunters-Canada/VaxFinder-Frontend/pull/26

PR #25, opened 2021-05-01
Codegen a typesafe API client

johnhok:
Generates typesafe React hooks using https://github.com/contiamo/restful-react based off the OpenAPI schema.

Also tried to simplify the readme.

Closes:
https://github.com/Vaccine-Hunters-Canada/VaxFinder-Frontend/issues/21


johnhok:
@pleckey do you happen to know why the build is failing here?

```
azure_static_web_apps_api_token was not provided.
The azure_static_web_apps_api_token is required for deploying content. If you'd like to continue the workflow run without deployment, add the workflow configuration skip_deploy_on_missing_secrets set to true
```

pleckey:
@johnhok this is because it is coming from a forked repo, and your fork (where the build is happening) doesn't have the deploy token.  You could either open the PR as a branch on the main repo, or setup your own deploy token with your own Azure account if you want to work off a fork.

pleckey:
@johnhok I will also look into disabling this step for forks, since that is a valid workflow - just not one I considered until now :)

src/App.tsx
```diff
@@ -74,9 +75,11 @@ export function App() {
 
   return (
     <AppProvider theme={theme} i18n={enTranslations}>
-      <Frame topBar={topBarMarkup} navigation={navigationMarkup}>
-        {getCurrentMarkup()}
-      </Frame>
+      <RestfulProvider base={process.env.REACT_APP_API_URL ?? ""}>
```
chrisahardie, line 78: Assigning an empty string will cause an unexpected and somewhat non-obvious error when we attempt to fetch, will it not? If we are missing the env var, it would probably be better to throw an explicit error detailing the cause, show a helpful message, and/or log it.

package.json
```diff
@@ -21,7 +22,9 @@
     "eject": "react-scripts eject",
     "lint": "eslint --max-warnings 0 --ext .tsx,.ts ./src",
     "lint:fix": "eslint --fix --ext .tsx,.ts,.js,.jsx ./src",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "generate:api": "restful-react import --file schema.json --output ./src/apiClient.tsx",
+    "postgenerate:api": "eslint --fix ./src/apiClient.tsx"
```
chrisahardie, line 27: Hmm, my take is we should probably just add `apiClient.tsx` to an `.eslintignore` file - it's a generated file, no point forcing generated code conform to our lint standards.

johnhok: It is somewhat useful to have it formatted consistently though as a way to make reading through the file easier. The library itself does suggest this post step.

src/App.tsx
johnhok: I was thinking this would default to the production API once available? But if that doesn't make sense I agree that it should throw an explicit error.

package.json
chrisahardie: > It is somewhat useful to have it formatted consistently though as a way to make reading through the file easier. The library itself does suggest this post step.

Fair enough, keeps the codebase from becoming inconsistent :)

johnhok:
I'm going to close this PR and open from a non-forked repo due to those deploy issues.

PR #28, opened 2021-05-02
23 consume availability endpoint

chrisahardie:


chrisahardie:
Opening a draft PR to solicit feedback - DO NOT MERGE

src/components/PharmacyList/PharmacyList.tsx
```diff
@@ -12,44 +13,83 @@ interface Pharmacy {
...
+      return {
+        id: pharmacy.id,
+        pharmacyName: pharmacy.location.name,
+        booking: true,
+        address: addressSegments.join(" "),
+        lastUpdated: "N/A",
+        phone: pharmacy.location.phone || "",
+        website: pharmacy.location.url || "",
+      };
+    }) || [];
```
johnhok, line 53: What do you think about not doing these prop transformations here and just passing the vaccine availability entry as a whole? The transformations here seem extraneous and for something like the address segments, that could just be handled in JSX? I think it may keep the types more simple since we don't need the `Pharmacy` interface?

src/apiClient.tsx
```diff
@@ -12,361 +12,379 @@ import {
...
 ) =>
   useGet<
-    EntriesResponse[],
+    VaccineAvailabilityExpandedResponse[],
     HTTPValidationError,
-    GetEntriesApiV1EntriesGetQueryParams,
+    ListVaccineAvailabilityApiV1VaccineAvailabilityGetQueryParams,
     void
-  >("/api/v1/entries/", props);
+  >("/api/v1/vaccine-availability", props);
```
picuron, line 147: Remove the props, as we are not using postal code just yet.
Should be `("/api/v1/vaccine-availability")` for now

picuron:
In .env, REACT_APP_API_URL should now be `http://vax-availability-api.azurewebsites.net`

src/apiClient.tsx
```diff
@@ -12,361 +12,379 @@ import {
...
+export interface RequirementResponse {
+  id: number;
+  name: string;
   description: string;
   created_at: string;
 }
 
+export interface VaccineAvailabilityExpandedResponse {
+  id: string;
+  numberAvaliable: number;
```
picuron, line 79: numberAvailable is spelled wrong here, which leads to it being undefined. Changing this to correct spelling resolved it.

src/mocks/vaccineAvailability/handlers.ts
```diff
@@ -0,0 +1,45 @@
...
+import { VaccineAvailabilityExpandedResponse } from "../../apiClient";
+
+export const vaccineAvailabilityHandlers = [
+  rest.get(
+    "http://localhost:5000/api/v1/vaccine-availability",
+    (req, res, ctx) => {
+      const response: VaccineAvailabilityExpandedResponse[] = [
+        {
+          id: "3fa85f64-5717-4562-b3fc-2c963f66afa6",
+          numberAvaliable: 4,
```
picuron, line 11: Need to change spelling here as well (or maybe we get rid of the mock files since we can get data now?)

src/apiClient.tsx
```diff
@@ -12,361 +12,379 @@ import {
...
   postcode: string;
-  latitude: number;
-  longitude: number;
-  geohash: string;
   created_at: string;
 }
 
-export interface EntriesResponse {
-  id: number;
-  numberAvaliable: number;
```
picuron, line 31: numberAvailable spelling here

src/components/PharmacyList/PharmacyList.tsx
chrisahardie: @johnhok I try to separate components that know about things external to themselves, and components that don't. It's basically the "smart/dumb" or "container/presentational" philosophy. As such, I don't want o give the entry directly to the pharmacy component as it is an API type, and then we have coupled a presentational component to our API. There is also more information in there than the component needs - it is better to be explicit about what exactly our component requires.

I did use the address segments to ensure spaces are added correctly - yes, it is probably overkill but I am OCD about my markup :) 

I did take your advice and dropped the pre-existing `Pharmacy` interface. It is better to infer props from the component you are pushing props into and supply them.

I've also re-named `PharmacyContainer` to `PharmacyCard` as "container" has a loaded meaning in React/Redux - it is a component that is aware of external state, which the component definitely should not.

johnhok: The new `PharmacyProps` typing is great. And those are great points. It would also make unit testing these components much easier without having to mock the whole API type.

package.json
```diff
@@ -8,6 +8,7 @@
     "@testing-library/jest-dom": "^4.2.4",
     "@testing-library/react": "^9.3.2",
     "@testing-library/user-event": "^7.1.2",
+    "cross-env": "^7.0.3",
```
johnhok, line 11: Should this be a dev dependency?

src/mocks/vaccineAvailability/handlers.ts
chrisahardie: Mocks continue to be useful to easily supply different sets of data (hey, let's throw a thousand items into the page and see if anything blows up!) and we will also use these in our tests. Unfortunately there is going to be maintenance work, but I think we should really keep these up to date.

package.json
chrisahardie: Oooh, yes... not that the distinction really matters, the build actually walks the graph via code, and doesn't refer to `package.json` at all. But for the sake of my ocd i will move.

src/components/PharmacyList/PharmacyList.tsx
```diff
@@ -1,61 +1,69 @@
...
-  pharmacyName: string;
-  booking: boolean;
-  address: string;
-  phone: string;
-  website: string;
-  lastUpdated: string;
-}
+type PharmacyProps = React.ComponentProps<typeof PharmacyCard>;
 
 export function PharmacyList() {
```
johnhok, line 7: Is this a good candidate to be a container? Might be useful to build out a `containers` and maybe even `routes` folder. I think there was also some tasks in the future to have an FAQ page so could be useful to scaffold out the foundation for this pattern. It may be out of the scope of this PR though.

package.json
chrisahardie: Done

johnhok (approved):

src/components/PharmacyList/PharmacyList.tsx
chrisahardie: Yes, was thinking the same thing, but yes, out of scope... this ticket was just to provide guidance about consuming an endpoint lol

PR #37, opened 2021-05-03
'as of' date-time formatting

shapesandobjects:


package.json
```diff
@@ -8,8 +8,10 @@
     "@testing-library/jest-dom": "^4.2.4",
     "@testing-library/react": "^9.3.2",
     "@testing-library/user-event": "^7.1.2",
+    "moment": "^2.29.1",
```
chrisahardie, line 11: There are better alternatives to moment, even the authors suggest looking for something else: https://momentjs.com/docs/#/-project-status/

johnhok (commented):
Screen Shot 2021-05-03 at 4 16 37 PM

A couple of suggestions.
1. Think we should display on the 12 hour clock for readability.
2. We should probably display a timezone if this is going to be viewed by people across the region so it's clear what timezone we're showing the times in?

package.json
chrisahardie: Yeah, it's def lightweight - moment is actually a MASSIVE dependency

chrisahardie:
> Screen Shot 2021-05-03 at 4 16 37 PM
> 
> A couple of suggestions.
> 
> 1. Think we should display on the 12 hour clock for readability.
> 2. We should probably display a timezone if this is going to be viewed by people across the region so it's clear what timezone we're showing the times in?

Yeah, the 24 hour clock and `PM` don't work together. 

Sigh... I was wondering how long it would be before we had to get into a conversation about timezones :)

picuron:
> > Screen Shot 2021-05-03 at 4 16 37 PM
> > A couple of suggestions.
> > 
> > 1. Think we should display on the 12 hour clock for readability.
> > 2. We should probably display a timezone if this is going to be viewed by people across the region so it's clear what timezone we're showing the times in?
> 
> Yeah, the 24 hour clock and `PM` don't work together.
> 
> Sigh... I was wondering how long it would be before we had to get into a conversation about timezones :)

picuron:
Wrong button, sorry ^

picuron:
And yeah I didnt even think of timezones. Does `date-fns` do any timezone conversions based on location, or would something else have to be used for that?

chrisahardie:
> And yeah I didnt even think of timezones. Does `date-fns` do any timezone conversions based on location, or would something else have to be used for that?

Yeah. there's an extension for that. We would read in the browser's timezone and then make the calculation to display the correct time with the timezone name - I am assuming the backend is storing in UTC? I'll confirm.

shapesandobjects:
Re-working this using date-fns-tz, so we'll have localized time zone formatting, etc.
Only hold-up atm is that our backend is returning naive datetime objects, which can be local time, not utc specific.  I don't know if this means client or server local time, so will get a PR into the backend to send as utc, and we can be confident of appropriate format handling on the frontend

chrisahardie:
> Re-working this using date-fns-tz, so we'll have localized time zone formatting, etc.
> Only hold-up atm is that our backend is returning naive datetime objects, which can be local time, not utc specific. I don't know if this means client or server local time, so will get a PR into the backend to send as utc, and we can be confident of appropriate format handling on the frontend

Awesome, that's the ticket - remove any and all ambiguity by getting a date with TZ or offset. 👍 👍 👍

shapesandobjects:
Updated to use `date-fns-tz`.  Force pushed to overwrite it all from scratch.

From testing with mock data, I can confirm that this adjusts a UTC time to the user's local system time.  Until the backend gets updated, there might be wonky times shown because of the half-fineshed end-to-end fix, but this half should be done.

chrisahardie (commented):
Could I ask that you improve a couple of my tests while you're touching this? We can now assert we are properly displaying dates in `PharmacyList.test.tsx`

Just ensure we can find the datetime in the proper format in the first couple tests:

```
...
await screen.findByRole("heading", { name: /stittsville ida/i });
await screen.findByText(/appointments available/i);
await screen.findByText(/as of may 1 2021, 5:07 pm edt/i);
...
```

```
await screen.findByRole("heading", { name: /stittsville ida/i });
await screen.findByText(/appointments not available/i);
await screen.findByText(/as of may 1 2021, 5:07 pm edt/i);
```

Is it weird to anyone else that the month isn't capitalized, and the timezone isn't uppercase?

johnhok (approved):
Looks great! 🔥

chrisahardie:
@shapesandobjects Sorry, seems like I made your life harder - tests failed, the markup being rendered is actually `May 1 2021, 9:07 PM UTC`

shapesandobjects:
> @shapesandobjects Sorry, seems like I made your life harder - tests failed, the markup being rendered on the CI server is actually `May 1 2021, 9:07 PM UTC`
> 
> I think this should do it:
> 
> ```
> await screen.findByText(/as of may 1 2021, 9:07 pm utc/i);
> ```
> 
> You can run the tests quickly on the command line with `yarn test Pharmacy` to make sure I haven't flubbed it again :)

Yeah, caught that.  I've updated them to (hopefully) be timezone agnostic so they'll pass for automated testing & any of us regardless of location

src/containers/PharmacyList/PharmacyList.test.tsx
```diff
@@ -4,13 +4,20 @@ import { VaccineAvailabilityExpandedResponse } from "../../apiClient";
 import { server } from "../../mocks/server";
 import { render, screen } from "../../testUtils";
 import { PharmacyList } from "./PharmacyList";
+import { format } from "date-fns-tz";
+
+const formattedDate = format(
```
chrisahardie, line 9: Awesome, now we'll get consistent results irrespective of the timezone of the machine the test is being run on

👍

chrisahardie (approved):

Issue #41, opened 2021-05-03
Implement router for application


johnhok:
We should implement a router for the application so we can support standard routing between pages and have better deep linking support for various flows.

Recommend that we utilize: https://reactrouter.com/web/guides/quick-start

chrisahardie:
👍 for this

To my mind, we would start with (at least) a couple routes:

`/` - shows form to submit postal code
`/search/:postalCode` - reads in postal code route param and performs the search

This will allow us to deep link directly to a search results screen for a particular postal code, so a direct URL could be given to someone, and the results screen could be bookmarked by the user for ease of use.

Thoughts @picuron @pleckey ? I know there were thoughts about using localStorage so people don't need to re-enter their postal code, but a bookmarked page might be another solution.

pleckey:
I can’t 👍 this enough. Yes to deep linkable routes. LocalStorage was more
for the eligibility criteria and not postcode but we can always add a “Save
these criteria” checkbox to the criteria view, and prompt on these routes
if none found, to enable deep links which are a huge plus.

chrisahardie:
Note to whomever implements this - please add a `MemoryRouter` to `AllTheProviders` in `testUtils.tsx` so our tests will continue to run :) Reach out if you have questions.

Issue #23, opened 2021-05-01
API Call Upon Postal Code Entry


picuron:
Using the postal code obtained in #22 , send this as a parameter as an API call for locations. Upon the received of these locations, pass this through PharmacyList's return.

PR #42, opened 2021-05-04
Utilize react router to handle routing

johnhok:
This PR implements a client side router into the application and refactors some current components to follow a more organized folder structure for routing. As part of implementing the client side router, I've also wired up the form to submit the postal code for the search page.

## Misc Notes

* removed `package-lock.json`.. we should only have one of these between `yarn.lock`
* `customRender` has been updated to return a `history` property that can be used in test assertions.. although I'm not sure if there may be a better way to do this.. would like some feedback on this

## Closes

https://github.com/Vaccine-Hunters-Canada/VaxFinder-Frontend/issues/41
https://github.com/Vaccine-Hunters-Canada/VaxFinder-Frontend/issues/23


johnhok:
So the routing works on the client side... but I think we may need to update the static server that is serving the UI to be able to pass these routes through to our SPA's client routing (https://create-react-app.dev/docs/deployment#serving-apps-with-client-side-routing).

@pleckey happen to have any idea what needs to be done here?

chrisahardie:
> So the routing works on the client side... but I think we may need to update the static server that is serving the UI to be able to pass these routes through to our SPA's client routing (https://create-react-app.dev/docs/deployment#serving-apps-with-client-side-routing).
> 
> @pleckey happen to have any idea what needs to be done here?

Oh, we need a url re-write rule. Basically every request needs to be redirected transparently to the root of the app, then the router kicks in, looks at the path and renders all the magic.

src/containers/Home/Home.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+    const { history } = render(<Home />);
+
+    const input = await screen.findByRole("textbox", {
+      name: /please enter your postal code/i,
+    });
+    fireEvent.change(input, { target: { value: "K2T0E5" } });
+    expect(input.value).toBe("K2T0E5");
+
+    const submit = screen.getByRole("button", { name: /submit/i });
+    fireEvent.click(submit);
```
chrisahardie, line 16: Minor, you should move to using `user-event` for interacting with dom nodes: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-testing-libraryuser-event

pleckey:
Please add the following as `routes.json` in the root of the output/build folder:

```
{
  "routes": [
    {
      "route": "/*",
      "serve": "/index.html",
      "statusCode": 200
    }
  ]
}
```

src/routes/Search.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+import { useParams } from "react-router-dom";
+import { PharmacyList } from "../containers/PharmacyList";
+
+interface SearchRouteParam {
+  postalCode: string;
+}
+
+// eslint-disable-next-line import/no-default-export
```
chrisahardie, line 9: Why are you disabling this? We should either follow the rule or we disable it. My issue with default exports is they can be named arbitrarily while imported. Named exports must be imported without ambiguity.

johnhok: React lazy requires default exports.

> React.lazy takes a function that must call a dynamic import(). This must return a Promise which resolves to a module with a default export containing a React component.

pleckey:
Please add the following so it outputs to `build/staticwebapp.config.json`:

```
{
  "navigationFallback": {
    "rewrite": "/index.html",
    "exclude": ["/*.{png,jpg,gif,ico}", "/static/*"]
  },
  "globalHeaders": {
    "content-security-policy": "default-src https: 'unsafe-eval' 'unsafe-inline'; object-src 'none'"
  }
}
```

this should allow react router to handle the routes

src/routes/Search.tsx
chrisahardie: dang it... i keep forgetting about that

johnhok:
> Please add the following so it outputs to `build/staticwebapp.config.json`:
> 
> ```
> {
>   "navigationFallback": {
>     "rewrite": "/index.html",
>     "exclude": ["/*.{png,jpg,gif,ico}", "/static/*"]
>   },
>   "globalHeaders": {
>     "content-security-policy": "default-src https: 'unsafe-eval' 'unsafe-inline'; object-src 'none'"
>   }
> }
> ```
> 
> this should allow react router to handle the routes

Worked perfectly, thanks!

https://witty-ocean-02e42dd0f-42.eastus2.azurestaticapps.net/search/k2T0E5

package.json
```diff
@@ -41,6 +42,8 @@
     "@testing-library/react": "^11.2.6",
     "@testing-library/user-event": "^13.1.8",
     "@types/jest": "^26.0.23",
+    "@types/react": "^17.0.4",
```
chrisahardie, line 45: we have a type mismatch with the version of react we are using.

src/containers/Home/Home.test.tsx
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+import { render, screen } from "../../testUtils";
+import { Home } from "./Home";
+import userEvent from "@testing-library/user-event";
+
+describe("Home", () => {
+  test("Should allow user to enter and submit valid postal code", async () => {
```
chrisahardie, line 7: More tests - huzzah!

package.json
johnhok: Why doesn't yarn add the correct version of types based off the dependency? :o

chrisahardie (approved):

chrisahardie:
> Please add the following so it outputs to `build/staticwebapp.config.json`:
> 
> ```
> {
>   "navigationFallback": {
>     "rewrite": "/index.html",
>     "exclude": ["/*.{png,jpg,gif,ico}", "/static/*"]
>   },
>   "globalHeaders": {
>     "content-security-policy": "default-src https: 'unsafe-eval' 'unsafe-inline'; object-src 'none'"
>   }
> }
> ```
> 
> this should allow react router to handle the routes

@johnhok If you have a chance, change `/static` to `public`, we aren't using it now but could in future.

pleckey:
> > Please add the following so it outputs to `build/staticwebapp.config.json`:
> > ```
> > {
> >   "navigationFallback": {
> >     "rewrite": "/index.html",
> >     "exclude": ["/*.{png,jpg,gif,ico}", "/static/*"]
> >   },
> >   "globalHeaders": {
> >     "content-security-policy": "default-src https: 'unsafe-eval' 'unsafe-inline'; object-src 'none'"
> >   }
> > }
> > ```
> > 
> > 
> > this should allow react router to handle the routes
> 
> @johnhok If you have a chance, change `/static` to `/public`, we aren't using it now but could in future.

May be best to just add `public` as `static` is used currently and this would break if `static` is removed:
image

chrisahardie:
> > > Please add the following so it outputs to `build/staticwebapp.config.json`:
> > > ```
> > > {
> > >   "navigationFallback": {
> > >     "rewrite": "/index.html",
> > >     "exclude": ["/*.{png,jpg,gif,ico}", "/static/*"]
> > >   },
> > >   "globalHeaders": {
> > >     "content-security-policy": "default-src https: 'unsafe-eval' 'unsafe-inline'; object-src 'none'"
> > >   }
> > > }
> > > ```
> > > 
> > > 
> > > this should allow react router to handle the routes
> > 
> > 
> > @johnhok If you have a chance, change `/static` to `/public`, we aren't using it now but could in future.
> 
> May be best to just add `public` as `static` is used currently and this would break if `static` is removed:
> image
> 
> @chrisahardie

Oh! CRA generally uses /public for static assets. Anyway, I have children who are going to jump on me in 7ish hours so I'll sign off. Don't hold up merging, we can attend to any issues as they arise.

pleckey (approved):

PR #45, opened 2021-05-04
Featured Community Projects Page

### LINKED ISSUE #39: Create Civic Projects Page
A page featuring links to other COVID projects that have been built by the community

Nicholas-Chong:
This PR contains the Community Projects page, featuring some of the projects that have been shared on the HackMD, VHC Twitter, and VHC Discord. I pretty much just winged the UI design so pls provide feedback :)

https://hackmd.io/@patcon/vax-app#-Related-Civic-Projects
closes #39 

## Other Things

1a960f249e0a80796301149a6997db212f190a36 @johnhok A small update to how routes work.  Makes container components return their children wrapped in a Polaris Page component. Now if we choose to add more pages, the page title can easily be set in the container component's code. See https://github.com/Vaccine-Hunters-Canada/VaxFinder-Frontend/blob/featured-civic-projects/src/containers/CommunityProjects/CommunityProjects.tsx#L29 for an example.

bea927fad713983e3e2d17e2648c56d3735d27f6 Had to add a {exactMatch: true} to the Home nav button because Polaris Navigation was highlighting Home even when it was inactive. Not really sure why this was happening but it seems to be fixed for now.

image

image


picuron:
**Please don't merge this yet.**

src/containers/CommunityProjects/CommunityProjects.tsx
```diff
@@ -0,0 +1,34 @@
...
+    );
+    infoCards.push(card);
+  });
+
+  return (
+    <Page title="Community Projects">
+      <p>Explore more COVID-19 projects built by people from across Canada!</p>
+      <div className="cardFlexBox">{infoCards}</div>
+    </Page>
+  );
```
johnhok, line 33: My suggestion is to just map over `projInfo` directly. I'm not a fan with interim variables and think JSX is well suited to handle these type of scenarios. You could also take it one step further and create a `CommunityProjectCard` component that could encapsulate the logic for rendering one of these cards.

```suggestion
  return (
    <Page title="Community Projects">
      <p>Explore more COVID-19 projects built by people from across Canada!</p>
      <div className="cardFlexBox">
        {projInfo.map((project, index) => (
          <div key={index} className="singleCardWrapper">
            <Card
              title={project.title}
              sectioned
              footerActionAlignment="left"
              primaryFooterAction={{
                content: "View Project",
                external: true,
                url: String(project.link),
              }}
            >
              <p>{project.description}</p>
            </Card>
          </div>
        )}
      </div>
    </Page>
  );
```

src/containers/CommunityProjects/CommunityProjectsInfo.tsx
```diff
@@ -0,0 +1,35 @@
+export const projInfo = [
```
johnhok, line 1: nit: prefer non abbreviations.. few extra characters but helps code readability 
```suggestion
export const projectInfo = [
```

src/containers/CommunityProjects/CommunityProjectsStyles.css
```diff
@@ -0,0 +1,42 @@
+.cardFlexBox {
```
johnhok, line 1: Not familiar with how CRA handles these styles. Are these styled scoped in any way? 

It would be great to introduce styled components into this project but this is out of scope of this PR.

src/containers/Home/Home.tsx
```diff
@@ -21,7 +28,7 @@ export function Home() {
     : undefined;
 
   return (
-    <>
+    <Page>
```
johnhok, line 31: 👍

src/containers/CommunityProjects/CommunityProjects.tsx
Nicholas-Chong: You're right this looks much better. I went ahead and created a separate component for the individual card as well. https://github.com/Vaccine-Hunters-Canada/VaxFinder-Frontend/pull/45/commits/e65f549e9bfee3d175a7ea42bd8fa2ae43e3477d
https://github.com/Vaccine-Hunters-Canada/VaxFinder-Frontend/pull/45/commits/2240bccb8b2b056e59857a39320a88e34e69bfea

src/containers/CommunityProjects/CommunityProjectsStyles.css
Nicholas-Chong: Yeah I also thought about using styled-components but didn't want to introduce any big changes. I didn't see any styling conflicts in the build as of right now, so the scope issue sorta went over my head.

One way to ensure there's no styling conflicts could be to just give the parent div an id and use it as a specific selector for all the appropriate styles (but this seems less than ideal)

johnhok: Take a look at CSS modules.

https://create-react-app.dev/docs/adding-a-css-modules-stylesheet/

It looks like it may be built into create react app. This would scope the CSS to prevent global clashing.

src/containers/CommunityProjects/CommunityProjectsStyles.css
```diff
@@ -0,0 +1,42 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  align-items: flex-start;
+  margin-top: 2em;
+}
+
+.singleCardWrapper {
+  width: 30%;
+  min-width: 216px;
+  margin-right: 1.5em;
```
shapesandobjects, line 11: we should think in `rem` rather than `em` or `px`, to accommodate people zooming for visibility, etc

src/containers/CommunityProjects/CommunityProjectsStyles.css
```diff
@@ -0,0 +1,42 @@
...
+    margin-right: 0;
+  }
+
+  .singleCardWrapper:last-child {
+    margin-bottom: 0;
+  }
+
+  /* Override Polaris Page CSS for mobile screens */
+  .Polaris-Page {
+    margin: 0 2vw 0 2vw;
```
shapesandobjects, line 31: just `margin: 0 2vw;` works.  Not a must-fix

src/containers/CommunityProjects/CommunityProjectsStyles.css
Nicholas-Chong: Yeah your right. Changed in 3e10178d5132a0d8c9db69e0bfc46944bbfd07d6

src/containers/CommunityProjects/CommunityProjectsStyles.css
Nicholas-Chong: Alright so I used the module.css and it worked pretty well. I still had to wrap the entire page in a communityProjectPage class. Since Polaris doesn't let you pass custom styles, this is the only way I could see to target Polaris styles without the css changes clashing with the Global,

55a4d9d

src/containers/CommunityProjects/CommunityProjectsInfo.tsx
Nicholas-Chong: 👍 bd33acc

src/components/CommunityProjectsCard/CommunityProjectCard.tsx
```diff
@@ -0,0 +1,22 @@
+import React from "react";
+import { Card } from "@shopify/polaris";
+import styles from "./CommunityProjectsStyles.module.css";
+
+export function CommunityProjectCard({ title, link, description }) {
```
chrisahardie, line 5: Why isn't this causing a compilation error? This is an implicit any...

src/components/CommunityProjectsCard/CommunityProjectCard.tsx
```diff
@@ -0,0 +1,22 @@
...
+  return (
+    <div className={styles.singleCardWrapper}>
+      <Card
+        title={String(title)}
```
chrisahardie, line 9: Once we have a props interface, TS will know `title` is a string and we won't need to wrap it in `String()`

src/components/CommunityProjectsCard/CommunityProjectsStyles.module.css
```diff
@@ -0,0 +1,26 @@
+.singleCardWrapper {
```
chrisahardie, line 1: Whoa, I had no idea this was a thingy in CRA!

src/components/CommunityProjectsCard/CommunityProjectCard.tsx
Nicholas-Chong: Ahh ok I understand I'll try and do that. In general, should I be creating a props interface for all components? Passing the interface like ```(props: PropsInterface)``` should avoid the implicit any issue as well right? I'm a huge typescript noob sorry lol

src/components/CommunityProjectsCard/CommunityProjectsStyles.module.css
Nicholas-Chong: Yeahhh its pretty cool. johnhok shared it with me.

src/components/CommunityProjectsCard/CommunityProjectCard.tsx
chrisahardie: Any component that accepts props should have those props typed to an interface so as to prevent vars with the wrong datatypes from getting passed in.

So the common pattern is

```
interface Props {
  title: string;
  ...
}

export const MyComponent(props: Props) => {
...
```

Nicholas-Chong: Got it. Added a props interface in 173c6ba

pleckey:
Temporarily closing this so we can open some staging slots.  Can be re-opened when comments are addressed and updated.

PR #54, opened 2021-05-05
51 updated api client

chrisahardie:


src/apiClient.tsx
```diff
@@ -59,31 +110,117 @@ export interface OrganizationCreateRequest {
...
+  numberAvailable?: number;
   numberTotal?: number;
-  date: string;
+  date?: string;
   vaccine?: number;
-  inputType: InputTypeEnum;
+  inputType?: InputTypeEnum;
   tags?: string;
-  created_at: string;
+  id: string | number;
```
chrisahardie, line 163: IMPORTANT NOTE TO WHOMEVER GENERATES THE API CLIENT: this id needs to be MANUALLY altered after client generation. `restful-react` types this to `{}` which is not correct. The source swagger JSON has this:

```
id:
          title: Id
          anyOf:
            - type: string
              format: uuid
            - type: integer
```
So basically this is an int or a guid, all good... But the `{}` that gets generated basically means any non-nullish value, which is wrong. Will file an issue with `restful-react`

src/mocks/vaccineAvailability/handlers.ts
```diff
@@ -15,6 +15,7 @@ export const vaccineAvailabilityHandlers = [
           inputType: 1,
           tags: "high-risk",
           created_at: "2021-05-01T21:07:28.232Z",
+          timeslots: [],
```
chrisahardie, line 18: Upcoming ticket to supply mock timeslot data should populate this

src/containers/PharmacyList/PharmacyList.tsx
```diff
@@ -19,7 +19,7 @@ export function PharmacyList(props: Props) {
     error,
   } = useListVaccineAvailabilityApiV1VaccineAvailabilityGet({
     queryParams: {
-      postalCode: props.postalCode,
+      postal_code: props.postalCode,
```
chrisahardie, line 22: Hrm, guess we went snake case on the back end.

src/apiClient.tsx
johnhok: This is pretty unfortunate. It looks somewhat related to this issue https://github.com/contiamo/restful-react/issues/317

Is there any way we can get the backend to return a single type for this ID? It seems like it should just be a string type like the other IDs being returned in the schema.

This manual step of updating the ID in the client here will be very easy to miss.

src/components/PharmacyCard/PharmacyCard.tsx
```diff
@@ -7,7 +7,7 @@ import Iframe from "react-iframe";
 interface PharmacyProps {
   // Id used for creating React keys
   // eslint-disable-next-line react/no-unused-prop-types
-  id: string;
+  id: string | number;
```
johnhok, line 10: Why do we need the ID here? I don't see it being used. The key prop that is set in the `PharmacyList` should be sufficient to setting the react key.

src/components/PharmacyList/PharmacyList.css
```diff
@@ -0,0 +1,6 @@
+.wrapper {
```
johnhok, line 1: Can we use CSS modules here to avoid global css clashing across containers and components.

src/components/PharmacyCard/PharmacyCard.tsx
chrisahardie: Well, the key prop is set in `PharmacyList` like this:

```
pharmacyList.map((pharmacy) => {
            return (
              <PharmacyCard
                  key={pharmacy.id}
```

Address could be used instead? Dunno if phone or website would be appropriate?

src/components/PharmacyList/PharmacyList.css
chrisahardie: Wait, this shouldn't have been included at all, what the heck

src/apiClient.tsx
chrisahardie: It won't be easy to miss - compilation will fail. It just won't be obvious as to how to resolve it, lol. Hrm, maybe a README update? Not perfect, but might help.

johnhok: Oh yeah, a failing build makes it more obvious. A readme update would probably help to resolve in the interim.

Still wondering why it's typed like this on the backend. Any insight @egamble01?

src/components/PharmacyCard/PharmacyCard.tsx
johnhok: Ah I see now.. since `pharmacyList` is depending on the props from `PharmacyCard`. I guess this is ok. My point was more about this interface having a prop that would be never used in the component but was there because a consumer depended on it.

The ID coming from the server is probably the most appropriate key here still I'd imagine.

johnhok (approved):
LGTM, left some replies, nothing blocking.

PR #57, opened 2021-05-06
Postal code format, as-of date handling

shapesandobjects:
Per a conversation w/ Sayon, the api is expecting only the first 3 characters of the postal code for searching.  These 3 chars should be lowercased for future caching consistency.
Also removed any middle space in order to maintain url consistency.

This should probably not be merged until after https://github.com/Vaccine-Hunters-Canada/VaxFinder-Frontend/pull/54

### DIFF
--- a/src/containers/Home/Home.tsx
+++ b/src/containers/Home/Home.tsx
@@ -10,7 +10,7 @@ export function Home() {
   const handleSubmit = () => {
     const postalCodeRegex = /[a-zA-Z][0-9][a-zA-Z](-| |)[0-9][a-zA-Z][0-9]/;
     if (postalCodeRegex.test(postalCode)) {
-      history.push(`/search/${postalCode}`);
+      history.push(`/search/${postalCode.toLowerCase().replace(" ", "")}`);
     } else {
       setShouldShowInvalidPostal(true);
     }
--- a/.env
+++ b/.env
@@ -1,1 +1,1 @@
-REACT_APP_API_URL=https://vax-availability-api.azurewebsites.net
+REACT_APP_API_URL=https://vax-availability-api-staging.azurewebsites.net
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "date-fns-tz": "^1.1.4",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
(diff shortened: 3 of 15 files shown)


chrisahardie:
> Per a conversation w/ Sayon, the api is expecting only the first 3 characters of the postal code for searching. These 3 chars should be lowercased for future caching consistency.
> Also removed any middle space in order to maintain url consistency.
> 
> This should probably not be merged until after #54

Sounds like there is some disconnect with what is happening with the api, the leads are jumping on a call tonight to sort it out - there may need to be another api client generated once the dust settles.

As an aside, a test broke as it asserted that there would be a 6-letter postal code in the querystring, so the test should be updated to expect only three now.

shapesandobjects:
--- DO NOT MERGE UNTIL POSTAL FORMATTING IS CONFIRMED ---

@chrisahardie I can't find the broken test - is it the one in `Home.test.tsx` that was expecting capitalization?

src/containers/Home/Home.tsx
```diff
@@ -10,7 +10,7 @@ export function Home() {
   const handleSubmit = () => {
     const postalCodeRegex = /[a-zA-Z][0-9][a-zA-Z](-| |)[0-9][a-zA-Z][0-9]/;
     if (postalCodeRegex.test(postalCode)) {
-      history.push(`/search/${postalCode}`);
+      history.push(`/search/${postalCode.toLowerCase().replace(" ", "")}`);
```
johnhok, line 13: I think that maybe this normalization is better done within the `PharmacyList` where the query param is being set in the API call? The problem is that users could still be hitting the search route directly and it would be nice to cover all those scenarios. Just a thought. Or perhaps it's better extracted to a utility and done in both areas to ensure consistency everywhere + consistent URLs that we're setting on the browser.

shapesandobjects: Good call.  I've added the lowercase & space removal there as well.  I'm going to leave this here still, as it's lightweight and gives us consistent urls.

I'm not going to worry about abstracting to a utility for this - it's lightweight enough to be doing a few js string functions and the manipulation is slightly different for both (i.e. full string vs only first three chars).

johnhok (approved):

pleckey (approved):

PR #59, opened 2021-05-06
Fix mobile layout alignment

Nicholas-Chong:
Fixes an issue where the cards and banner on the /search path are pushed too far to the right on mobile screens. Problem was being caused by nested <Layout.Section>'s

| Before (iPhone X)  | After (iPhone X) |
| ------------- | ------------- |
| image | image |

Something to note: Containers and individual Components should not return anything wrapped in `<Layout.Section>` since the AppFrame already wraps the Route's response in a `<Layout.Section>`

### DIFF
--- a/src/containers/PharmacyList/PharmacyList.tsx
+++ b/src/containers/PharmacyList/PharmacyList.tsx
@@ -102,27 +102,28 @@ export function PharmacyList(props: Props) {
     );
 
   return (
-    <section aria-label="pharmacy-list">
+    <>
       {shouldShowBanner ? (
         <EligibilityBanner onDismiss={() => setShouldShowBanner(false)} />
       ) : null}
-
-      {pharmacyList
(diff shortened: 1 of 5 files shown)


src/containers/PharmacyList/PharmacyList.tsx
```diff
@@ -104,27 +104,28 @@ export function PharmacyList(props: Props) {
...
-                lastUpdated={pharmacy.lastUpdated}
-                pharmacyName={pharmacy.pharmacyName}
-                phone={pharmacy.phone}
-                website={pharmacy.website}
-              />
-            );
-          })
-        : null}
-    </section>
+      <section aria-label="pharmacy-list" style={{ marginTop: "2rem" }}>
```
Nicholas-Chong, line 111: `2rem` is the same marginTop provided to all Polaris-Card except the first

PR #65, opened 2021-05-09
Add i18n to switch between English and French

andrewpmk:
(Recreating pull request from branch Vaccine-Hunters-Canada instead of my personal repo because I get Azure errors otherwise)

Hi all, my real name is Andrew MacKinnon and I'm a web developer at Forum Research (though most of my experience is with ASP.NET - sorry I'm new to React). I'm interested in contributing to this project because I'm trying to get vaccinated like everyone else, and I'm trying to improve my JavaScript skills.

I have added English and French translations using the library i18next (not sure if this is the best i18n library to use for this project). https://react.i18next.com/ https://www.i18next.com/ I have added a button on the left-hand side for switching between English and French. It seems to be working right now although the tests give warnings because the resource strings aren't being loaded in the tests (not sure how to fix this). I modified the strings that the tests are searching to the resource keys so that the tests pass (with warnings). Ideally we will add more languages besides English and French to encourage more people to get vaccinated!

In addition I removed junk left behind by create-react-app (the React logo and some CSS) that isn't being used.

What license is this project licensed under - there isn't a LICENSE file right now. MIT? GPL version 2? Public domain? I'm making these contributions public domain to avoid any uncertainty about licenses.


andrewpmk:
Should be fixed now. What was happening is that i18next is set up to autodetect language based on browser settings and it is set to "en-CA" (but could be set to en-US, en-GB, fr-CA, or many other possible values). But I was checking for the value "en" not "en-CA" in the language toggle button so it wasn't working properly. Set it to only look at the first 2 characters of the language code (the first part of the language code is always 2 characters right?)

The way this should work is if your browser language is French it sets default language to French. If it is set to English or any other language it should set default language to English. Language toggle switches between English and French. In the future we may support more languages if we can find someone to translate strings for us.

src/components/PharmacyCard/PharmacyCard.tsx
```diff
@@ -17,15 +20,26 @@ interface PharmacyProps {
 }
 
 export function PharmacyCard(props: PharmacyProps) {
+  const { t, i18n } = useTranslation();
   const [shouldShowMap, setShouldShowMap] = useState(false);
+  function FormatDate(date: Date) {
+    // alert(i18n.language);
```
shapesandobjects, line 26: We can remove the `alert`

src/components/PharmacyCard/PharmacyCard.tsx
```diff
@@ -17,15 +20,26 @@ interface PharmacyProps {
...
       return (
         <Banner status="success">
           <p>
-            <strong>Appointments available</strong>
+            <strong>{t("appointmentsavailable")}</strong>
             {props.lastUpdated.length > 0 && (
-              <> as of {format(new Date(props.lastUpdated), "MMM d, y")}</>
+              <>
+                {" "}
+                {t("asof")} {FormatDate(new Date(props.lastUpdated))}
```
shapesandobjects, line 41: We should use i18next's interpolation to include the formatted date as part of the translation string itself (https://www.i18next.com/translation-function/interpolation).
In general, we don't know if other language translations might have the date midway through the statement, so we shouldn't rely on tacking it on the end.

src/components/PharmacyCard/PharmacyCard.tsx
```diff
@@ -34,9 +48,12 @@ export function PharmacyCard(props: PharmacyProps) {
     return (
       <Banner status="critical">
         <p>
-          <strong>Appointments not available</strong>
+          <strong>{t("appointmentsnotavailable")}</strong>
           {props.lastUpdated.length > 0 && (
-            <> as of {format(new Date(props.lastUpdated), "MMM d, y")}</>
+            <>
+              {" "}
+              {t("asof")} {FormatDate(new Date(props.lastUpdated))}
```
shapesandobjects, line 55: see above re: intepolation

src/components/PharmacyCard/PharmacyCard.tsx
andrewpmk: Removed. Does anyone know how to get the JavaScript debugger working in Visual Studio Code with yarn start (I'm using Firefox)?

src/components/PharmacyCard/PharmacyCard.tsx
andrewpmk: Switched to using i18next interpolation, note I need to configure the date-fns-tz format function in the i18next configuration because i18next doesn't provide a date formatting library (it suggests moment but moment is outdated)

andrewpmk:
I have updated the pull request to make the following changes:
- Switched to using the string interpolation provided by i18next - note it does not provide built in date interpolation so that must be setup in i18n.ts using date-fns-tz
- Changed French translation of "as of" to "mis à jour le" - don't think "à partir du" makes sense, does anyone have a better translation?
- Merge changes from main ("No nearby pharmacies were found near {{postalCode}}. Please try again later." message) and translate
- Removed the useless alert

src/components/PharmacyCard/PharmacyCard.tsx
chrisahardie: I would typical only use the VS Code debugger during tests... `yarn start` spins up a server to serve the app to a browser, which has its own runtime, and own debugger. Why wouldn't you use FF's devtools?

andrewpmk:
Merged changes from main again. The changes in https://github.com/Vaccine-Hunters-Canada/VaxFinder-Frontend/pull/68 seem to have made my language toggle slow - maybe because it's downloading and displaying data for dozens of pharmacies in Toronto at once? Looks like we may need to display the pharmacies on pages instead of displaying all of them because there's 1000s of pharmacies in the GTA?

package.json
```diff
@@ -7,8 +7,13 @@
     "@shopify/polaris-icons": "^4.5.0",
     "date-fns": "^2.21.1",
     "date-fns-tz": "^1.1.4",
+    "i18next": "^20.2.2",
+    "i18next-browser-languagedetector": "^6.1.0",
+    "i18next-fs-backend": "^1.1.1",
```
chrisahardie, line 12: Don't think we need these two packages... they are for retrieving translations from a backend, which we are not doing. We are bundling our translations.

package.json
```diff
@@ -44,6 +49,7 @@
     "@testing-library/jest-dom": "^5.12.0",
     "@testing-library/react": "^11.2.6",
     "@testing-library/user-event": "^13.1.8",
+    "@types/i18next-fs-backend": "^1.0.0",
```
chrisahardie, line 52: Can be removed

src/AppFrame.tsx
```diff
@@ -1,15 +1,28 @@
...
 import { useLocation } from "react-router-dom";
 import { Frame, Layout, Navigation, Page, TopBar } from "@shopify/polaris";
 import { HomeMajor } from "@shopify/polaris-icons";
 import { Routes } from "./Routes";
 import { TwitterTimelineEmbed } from "react-twitter-embed";
+import { useTranslation } from "react-i18next";
+import i18next from "i18next";
 
 export function AppFrame() {
+  const { t, i18n } = useTranslation(undefined, { useSuspense: false });
```
chrisahardie, line 12: You can't just use `const { t, i18n } = useTranslation()`

src/AppFrame.tsx
```diff
@@ -1,15 +1,28 @@
...
   const location = useLocation();
   const [isMobileNavigationActive, setIsMobileNavigationActive] = useState(
     false,
   );
+  function languageToggle() {
+    // Display link to toggle to whichever language isn't set right now
+    // Get language code excluding country
+    // Initially language code set by browser so it is something like en-CA or en-US
+    if (i18n.language !== undefined && i18n.language.substring(0, 2) === "fr") {
```
chrisahardie, line 21: No action required, but do I remember hearing that we would support more than the two official languages? Maybe post-MVP? Should we create a ticket to update this toggle to a dropdown or something? cc @picuron

chrisahardie, line 21: Minor, could be shortened to `i18n.language?.substring(0, 2) === "fr"`

src/AppFrame.tsx
```diff
@@ -31,9 +44,22 @@ export function AppFrame() {
...
-                label: "Home",
+                label: t("home"),
                 icon: HomeMajor,
               },
+              {
+                label: languageToggle(),
+                onClick: () => {
+                  if (
+                    i18n.language !== undefined &&
+                    i18n.language.substring(0, 2) === "fr"
```
chrisahardie, line 55: Minor, could be shortened to `i18n.language?.substring(0, 2) === "fr"`

src/containers/Home/Home.test.tsx
```diff
@@ -28,7 +28,7 @@ describe("Home", () => {
     userEvent.type(input, "78701");
     expect(input.value).toBe("78701");
 
-    const submit = screen.getByRole("button", { name: /submit/i });
+    const submit = screen.getByRole("button", { name: /Submit/i });
```
chrisahardie, line 31: Minor, but this change isn't doing anything :) The `i` flag on the regex means it should match insensitively.

package.json
andrewpmk: i18next-browser-languagedetector detects the current language based off of a browser's language settings. Try changing the preferred language to different languages ("Choose your preferred language for displaying pages" in Firefox) and see what happens (use incognito mode). It should display French if language is set to fr-CA, fr-FR, etc., English if language is set to en-CA, en-US, etc. and English if any other language is set.

i18next-fs-backend is used in the tests to load the translation.json files, it is not used anywhere else.

package.json
andrewpmk: i18next-fs-backend is used in the tests to load the translation.json files.

src/AppFrame.tsx
andrewpmk: translating the Home button (Accueil) won't work without this. It gives an error if you remove useSuspense: false. Suspense is a react feature to show a loading spinner/image when data is loading. Alternatively if you wrap the AppFrame in <Suspense> tags it should work

src/AppFrame.tsx
andrewpmk: I know the toggle only supports two languages, but we need someone to translate some resource strings first :)

src/i18n.ts
```diff
@@ -0,0 +1,41 @@
...
+import enCA from "date-fns/locale/en-CA";
+import frCA from "date-fns/locale/fr-CA";
+// don't want to use this?
+// have a look at the Quick start guide
+// for passing in lng and translations on init
+
+i18n
+  // load translation using http -> see /public/locales (i.e. https://github.com/i18next/react-i18next/tree/master/example/react/public/locales)
+  // learn more: https://github.com/i18next/i18next-http-backend
+  .use(Backend)
```
chrisahardie, line 16: Oh, I see what you were doing - you were setting up an ajax call on our own web server. You don't need to do that, I would just include them directly... they are so small I don't see much benefit in lazy loading them.

src/i18n.ts
```diff
@@ -0,0 +1,41 @@
...
+  // detect user language
+  // learn more: https://github.com/i18next/i18next-browser-languageDetector
+  .use(LanguageDetector)
+  // pass the i18n instance to react-i18next.
+  .use(initReactI18next)
+  // init i18next
+  // for all options read: https://www.i18next.com/overview/configuration-options
+  .init({
+    fallbackLng: "en",
+    debug: true,
```
chrisahardie, line 26: We don't want to debug in prod, I would turn it off entirely, but if you want to keep it, we'll need to introduce an environmental var for it

src/AppFrame.tsx
andrewpmk: eslint doesn't complain :)

src/AppFrame.tsx
andrewpmk: eslint doesn't complain :)

src/containers/Home/Home.test.tsx
andrewpmk: I changed the tests to look for the resource key, then changed them back to look for the English string once I fixed the tests. Just ignore this change it doesn't matter

src/i18n.ts
```diff
@@ -0,0 +1,41 @@
...
+  // learn more: https://github.com/i18next/i18next-http-backend
+  .use(Backend)
+  // detect user language
+  // learn more: https://github.com/i18next/i18next-browser-languageDetector
+  .use(LanguageDetector)
+  // pass the i18n instance to react-i18next.
+  .use(initReactI18next)
+  // init i18next
+  // for all options read: https://www.i18next.com/overview/configuration-options
+  .init({
```
chrisahardie, line 24: Here is how I have set things up in the past to eager load the localizations:

```
import englishTranslations from "src/locales/en-CA.json";
import frenchTranslations from "src/locales/fr-CA.json";
...
const CONSTANTS = {
  ENGLISH_CODE: "en-CA",
  FRENCH_CODE: "en-FR",
}

const resources = {
  [CONSTANTS.ENGLISH_CODE]: {
    translation: englishTranslations,
  },
  [CONSTANTS.FRENCH_CODE]: {
    translation: frenchTranslations,
  },
};

i18next
  .use(detector)
  .use(initReactI18next)
  .init({
    resources,
    fallbackLng: CONSTANTS.ENGLISH_CODE,
  });
```

src/i18n.ts
```diff
@@ -0,0 +1,41 @@
...
+    fallbackLng: "en",
+    debug: true,
+
+    interpolation: {
+      escapeValue: false, // not needed for react as it escapes by default
+      // eslint-disable-next-line func-names
```
chrisahardie, line 30: Can't you just name the function rather than disabling the rule?

src/i18n.ts
```diff
@@ -0,0 +1,41 @@
...
+      format: function (value, format, lng) {
+        if (value instanceof Date && format !== undefined) {
```
chrisahardie, line 32: Nice! Can you not type the function you assign to `format` at all? If not, then you should evaluate `value instanceof Date && format` as you are only checking that `format` is `undefined` and not `null` or empty string.

src/i18n.ts
```diff
@@ -0,0 +1,41 @@
...
+          if (lng !== undefined && lng.substring(0, 2) === "fr") {
```
chrisahardie, line 33: `(lng?.substring(0, 2) === "fr")

src/testUtils.tsx
```diff
@@ -18,13 +20,18 @@ const customRender = (
       <Router history={history}>
         <AppProvider theme={theme} i18n={enTranslations}>
           <RestfulProvider base={process.env.REACT_APP_API_URL ?? ""}>
-            {children}
+            <Suspense fallback={<Spinner />}>
```
chrisahardie, line 23: Ah, maybe you do need to set `suspense` to false from earlier, but I think only if you grab your translations via ajax.

.jshintrc
```diff
@@ -0,0 +1 @@
+{"esversion": 9}
```
chrisahardie, line 1: a JSHint config is not required, we are using Typescript

andrewpmk: I put this in to stop VSCode from complaining should I remove it?

src/AppFrame.tsx
chrisahardie: 👍

chrisahardie: Nope, and this was the only way to evaluate your case for many years. But the modern way to do it is with the chaining operator as I have done above. Imagine if you were drilling through a chain of properties to evaluate a value... it is extremely cumbersome to check for existence on every property.

src/containers/Home/Home.test.tsx
chrisahardie: All good, you mentioned in the channel you were looking to improve your js, so I just wanted to mention this.

andrewpmk:
Updated pull request with some changes suggested by chrisahardie (except the eager loading of translations, I haven't changed this).

src/testUtils.tsx
andrewpmk: the tests will fail without the <Suspense>

chrisahardie (commented):
I think there might be merge conflicts if you merge from `main` ... we had a somewhat drastic change to the API, and those changes had to be cascaded to this web app as we need them to continue development on other tickets.

I owe you a 🍺 or seven, this is great work and sorely needs to be done! Thanks for pitching in on this! I'll float around to watch this PR and try to help get it through ASAP.

src/testUtils.tsx
andrewpmk: The tests will fail without the `<Suspense>` tag.

.jshintrc
chrisahardie: What was the complaint? Sounds like a workstation issue.

andrewpmk:
I merged two different commits made today into my pull request, there shouldn't be any more merge conflicts (unless you update main again).

.jshintrc
andrewpmk: removed

src/i18n.ts
andrewpmk: changed

src/i18n.ts
andrewpmk: Changed this to a named function

src/i18n.ts
andrewpmk: changed to a named function

src/i18n.ts
andrewpmk: Didn't notice that, just copied from the i18next config. Changed to false

src/i18n.ts
andrewpmk: The configuration on i18next help uses lazy loading and I copied from that. Is there a significant performance improvement if I change this?

package.json
chrisahardie: I meant `i18next-fs-backend` and `i18next-http-backend` both of which are used only to pull in our translations via AJAX instead of just adding them directly to our bundle. I don't see the need to pull in our translations via AJAX... It means we need to add two deps, our configuration is more complicated, and loading the translations is a bit slower as it means initiating an ajax request after our app is loaded.

src/i18n.ts
chrisahardie: Lazy loading means:

* more dependencies
* a more complicated configuration
* slightly longer TTI as we can't do anything until the web app spins up, initiates the request, receives the translation and applies them.

Lazy loading reaps benefits when the resources being loaded are large and varied. If our localization files were thousands of lines and spanned 100 languages, then lazy loading would absolutely be appropriate. But when we're talking about a couple tiny files, the network overhead of the AJAX requests to fetch them means there is no value in going this way.

chrisahardie:
> I merged two different commits made today into my pull request, there shouldn't be any more merge conflicts (unless you update main again).

👍

chrisahardie (approved):

chrisahardie (approved):
I'll approve, the eager-loading bit can come in a future ticket.

PR #70, opened 2021-05-09
-WIP- Don't merge: Added data table

picuron:


MattAlp (changes requested):
Regardless of the DNM status, we don't need to include the yarn files (lock, error log) and the other stuff we've been dragging into the repo. Let's add a `.gitignore`.

chrisahardie:
> Regardless of the DNM status, we don't need to include the yarn files (lock, error log) and the other stuff we've been dragging into the repo. Let's add a `.gitignore`.

Lock files should always be included in source control. They ensure consistent dependencies across installs. https://classic.yarnpkg.com/en/docs/yarn-lock/#toc-check-into-source-control

chrisahardie (commented):
Please delete `package-lock.json` - if we are using Yarn, we should not be using `npm` for package management.

src/components/PharmacyCard/PharmacyCard.tsx
```diff
@@ -52,6 +62,43 @@ export function PharmacyCard(props: PharmacyProps) {
       styles={{ border: "none" }}
     />
   );
+  const generateRows = () => {
```
chrisahardie, line 65: Could be reduced to with a more functional approach to something like:

```
const generateRows = () =>
    props.vaccineAvailabilities.map((availability) => [
      format(new Date(availability.date), "MMM d, y"),
      availability.numberAvailable,
    ]);
```

src/components/PharmacyCard/PharmacyCard.tsx
```diff
@@ -52,6 +62,43 @@ export function PharmacyCard(props: PharmacyProps) {
...
+    if (props.vaccineAvailabilities.length !== 0) {
+      return (
+        <Card.Section title="Dates">
+          <Button
+            primary
+            onClick={() => {
+              setShouldShowSlots(!shouldShowSlots);
+            }}
+          >
+            See Slots
```
chrisahardie, line 94: You need to toggle this text between "Show" and "Hide"

Now that we have localization rolled in, we should update the locale files... We'll also need a process to get the outstanding French translations done and rolled into the codebae.

src/components/PharmacyCard/PharmacyCard.tsx
```diff
@@ -1,9 +1,33 @@
...
+} from "@shopify/polaris";
 import { DomainsMajor, LocationMajor } from "@shopify/polaris-icons";
 import Iframe from "react-iframe";
+import {
+  VaccineAvailabilityTimeslotRequirementExpandedResponse,
+  VaccineAvailabilityRequirementsResponse,
+} from "../../apiClient";
 import { useTranslation } from "react-i18next";
 
+interface RequirementsWithAvailabilityInterface
```
chrisahardie, line 19: The naming convention suggested by Microsoft is to prefix interfaces with `I` like C# - but for whatever reason that never took off, lol.

src/components/PharmacyCard/PharmacyCard.tsx
```diff
@@ -1,9 +1,33 @@
...
+  extends VaccineAvailabilityRequirementsResponse {
+  numberAvailable: number;
+}
+
+interface VaccineAvailabilitiesByDateAndRequirementsInterface {
+  [key: string]: {
```
chrisahardie, line 25: Minor, but naming this member `date` instead of `key` would better express intent.

src/components/PharmacyCard/PharmacyCard.tsx
```diff
@@ -14,11 +38,18 @@ interface PharmacyProps {
   phone: string;
   website: string;
   lastUpdated: string;
+  vaccineAvailabilities: VaccineAvailabilityTimeslotRequirementExpandedResponse[];
+  vaccineAvailabilitiesNew: VaccineAvailabilitiesByDateAndRequirementsInterface;
 }
 
+// TODO: We don't need to carry both vaccineAvailabilities & vaccineAvailabilitiesNew,
```
chrisahardie, line 45: Get rid of the legacy one, if there are improvements required for the new one we can add spit and polish.

src/components/PharmacyCard/PharmacyCard.tsx
```diff
@@ -53,6 +84,61 @@ export function PharmacyCard(props: PharmacyProps) {
...
+      // props.vaccineAvailabilitiesNew[date].requirements.forEach(
+      //   (requirement) => {
+      //     rows.push([requirement.description, requirement.numberAvailable]);
+      //   },
+      // );
+    });
+
+    return rows;
+  };
+  const dataTableMarkup = shouldShowSlots ? (
```
chrisahardie, line 116: Are we going to break down amount by eligibility requirement? Or is that coming in a future ticket?

src/containers/PharmacyList/PharmacyList.tsx
```diff
@@ -123,16 +128,56 @@ export function PharmacyList(props: Props) {
...
+          return vaccineAvailability.numberAvailable > 0;
+        },
+      );
+
+      interface RequirementsWithAvailabilityInterface
+        extends VaccineAvailabilityRequirementsResponse {
+        numberAvailable: number;
+      }
+
+      interface VaccineAvailabilitiesByDateAndRequirementsInterface {
```
chrisahardie, line 143: This is duplicated from `PharmacyCard` - it is common for the parent component to need to know the shape of the props for the child component it is rendering. There are two approaches to tackle this:

Better: Export the interface from the child, and import it into the parent

Best: Have the parent infer the props of the child e.g. `type PharmacyCardProps = React.ComponentProps<typeof PharmacyCard>;`

src/utils/mock.ts
```diff
@@ -0,0 +1,416 @@
+const vaccineLocationOneAvailability = {
```
chrisahardie, line 1: I would name this file to something more precise, `mock` could be any number of things. And this shouldn't be in the `utils` folder, this is mock data, should be under `/mocks` or else it will be included in our deployment build.

chrisahardie (approved):

PR #72, opened 2021-05-10
postal code utilities

shapesandobjects:


src/utils/postalCode.ts
```diff
@@ -0,0 +1,29 @@
+const postalCodeIsValid = (postalCode: string) => {
+  return /[a-zA-Z][0-9][a-zA-Z][0-9][a-zA-Z][0-9]/.test(
```
plemarquand, line 2: This allows some invalid postal codes since letters that can be confused with numbers aren't allowed (O/0, I/1 etc). This is a stricter regex (also note the /i at the end for case insensitivity):

`/^[ABCEGHJ-NPRSTVXY]\d[ABCEGHJ-NPRSTV-Z][ -]?\d[ABCEGHJ-NPRSTV-Z]\d$/i`

andrewpmk: You might want to also flag ridiculous postal codes like H0H 0H0 as invalid.

Also note the first character of a postal code indicates the province: https://www150.statcan.gc.ca/n1/pub/92-195-x/2011001/other-autre/pc-cp/tbl/tbl9-eng.htm maybe we could make use of this somehow?

shapesandobjects: Updating to use the better postal code regex above, which also accounts for allowed first characters based on province.
Flagging for goofy postal codes could get tricky & inherently subjective.  Just getting a "no results" for any bad postal codes is probably sufficient for now

PR #75, opened 2021-05-13
74 regen api client

chrisahardie:


README.md
```diff
@@ -87,3 +87,22 @@ $ yarn build
...
+}
+```
+
+This will produce an error:
+
+> Don't use `{}` as a type. `{}` actually means "any non-nullish value".
+
+When this happens, you will need to examine the OpenApi schema and determine what the appropriate type should be, and manually update `apiClient.ts`
+
+Bug to track: https://github.com/contiamo/restful-react/issues/317
```
johnhok, line 108: 👍 

Although this bug is pretty unfortunate in `restful-react`, it's still not clear to me why a field like `takenAt` would be more than one type. The same with a UUID on an entity (which I think was what was using `anyOf` before). Is this just a mistyping from the backend or does this field legitimately support say a string and int type?

johnhok (approved):

PR #77, opened 2021-05-16
Support more languages

andrewpmk:
This pull request is a work in progress don't merge. I am working on adding support for more languages in addition to English and French. How should the controls for switching languages be designed? Should we use a dropdown menu or should we just put a large number of language links on the left-hand side (like Wikipedia?) Do we have any volunteers to translate into more languages? I may do some translating with Google Translate into a few languages for test purposes in the meantime.

I modified code to use eager loading as suggested by chrisahardie which should reduce load time if there are a small number of resource strings and languages. We may want to switch back to lazy loading if we add a lot of different languages though.


src/AppFrame.tsx
```diff
@@ -47,18 +112,13 @@ export function AppFrame() {
...
-                  if (i18n.language?.substring(0, 2) === "fr") {
-                    i18next.changeLanguage("en-CA");
-                  } else {
-                    i18next.changeLanguage("fr-CA");
-                  }
-                },
-              },
             ]}
           />
+          <Navigation.Section // Todo translating the word "Language" and displaying icon doesn't work for some reason
```
chrisahardie, line 117: `title={t("language")}` doesn't work?

src/AppFrame.tsx
```diff
@@ -7,22 +7,87 @@ import { Routes } from "./Routes";
...
+  /*
+  {
+    langCode: "zh-TW",
+    label: "繁體中文",
+  },
+  */
+];
+
+function LanguageBar(): ItemProps[] {
+  const output: ItemProps[] = languages.map((x) => {
```
chrisahardie, line 71: Minor, but my preference is for variables to have descriptive names, so instead of `x` I would use `language`

src/AppFrame.tsx
andrewpmk: I'm not sure why this doesn't work, seems to be a Shopify Polaris issue. Can you figure out a workaround? Also I was hoping to put a language icon here but that doesn't work either.

src/AppFrame.tsx
```diff
@@ -7,22 +7,87 @@ import { Routes } from "./Routes";
...
+  },
+  /*
+  {
+    langCode: "zh-TW",
+    label: "繁體中文",
+  },
+  */
+];
+
+function LanguageBar(): ItemProps[] {
```
chrisahardie, line 70: Another coding convention I like to follow is functions names generally have a verb prefix - e.g. `getLanguageBarItemProps` would give me a better sense of what this function is doing.

Of course now React components are also functions, and they are generally nouns :)

andrewpmk: changed

src/AppFrame.tsx
andrewpmk: changed

src/i18n.ts
```diff
@@ -1,40 +1,65 @@
...
+    fr: { translation: fr },
+    pl: { translation: pl },
+    ru: { translation: ru },
+    "zh-CN": { translation: zhCN },
+  },
+  interpolation: {
+    escapeValue: false, // not needed for react as it escapes by default
+    format: function format(value, fmt, lng) {
+      if (value instanceof Date && fmt !== undefined) {
+        const lang = lng?.substring(0, 2);
```
chrisahardie, line 36: Someone mentioned on the channel about supporting multiple Chinese languages - should we be evaluating the full string?

andrewpmk: There's two ways of writing Chinese - Simplified Chinese (zh-CN) and Traditional Chinese (zh-TW, zh-HK, zh-MO). Only Simplified Chinese is implemented for now because I haven't received translations for both. Date format currently being used is like 2021 5月 22 I think it's the same for both Simplified Chinese and Traditional Chinese (sorry I don't speak Chinese).

src/i18nfortests.ts
```diff
@@ -1,34 +1,16 @@
 /* eslint-disable import/no-default-export */
 import i18n from "i18next";
 import { initReactI18next } from "react-i18next";
-import Backend from "i18next-fs-backend";
-import { format as date_fnz_tz_format } from "date-fns-tz";
-import { enCA, frCA } from "date-fns/locale";
+import { i18nconfig } from "./i18n";
+import _ from "lodash";
+
+const i18nconfig_test = _.cloneDeep(i18nconfig); // Copy default configuration
```
chrisahardie, line 7: For one offs like this, I would generally just pull in the module you need instead of all of `lodash` - all you need is this: https://www.npmjs.com/package/lodash.clonedeep

As an alternative, when the object contains just simple values you can deep clone with `JSON.parse(JSON.stringify(i18nconfig))`

src/i18n.ts
andrewpmk: Simplified Chinese is predominantly used in Mainland China and Traditional Chinese is predominantly used in Hong Kong/Macau/Taiwan.

src/i18nfortests.ts
andrewpmk: I keep getting errors when I try to change this - "could not find declaration file for lodash.clonedeep" - won't fix

chrisahardie: You are introducing thousands of lines of code we aren't using. The lodash maintainers spent a non-trivial amount of time to [modularize lodash](https://www.npmjs.com/search?q=keywords:lodash-modularized) so as to reduce the surface area for bugs to exist on and to keep our bundle size as small as possible.

The error you posted is incomplete - it also includes:

> Try `npm i --save-dev @types/lodash.clonedeep` if it exists or add a new declaration (.d.ts) file containing `declare module 'lodash.clonedeep';`

Run `yarn @types/lodash.clonedeep` and the type will be pulled into the codebase. Note you may need to restart your IDE.

src/AppFrame.tsx
chrisahardie: Localizing the text works for me:

image

src/i18nfortests.ts
andrewpmk: Changed to lodash.clonedeep and installed @types/lodash.clonedeep and it works now. Don't think this really matters though. Shopify Polaris lists all of lodash as a dependency.

chrisahardie: Yes it does. Node modules are installed as a tree hierarchy. Multiple versions of the same dependency can exist in the tree.  Full Lodash weighs in at 75kb, so it's not a tiny dependency. 

https://lexi-lambda.github.io/blog/2016/08/24/understanding-the-npm-dependency-model/

src/AppFrame.tsx
andrewpmk: Fixed. It won't let me put an icon in though which I wanted to do

src/i18nfortests.ts
andrewpmk: It's changed to lodash.clonedeep - it won't let me put this in devDependencies even though I'm using this solely for tests (to clone i18n configuration and change default language to en instead of using the language detector). I think webpack may be smart enough to remove lodash in the build but not sure of this.

src/AppFrame.tsx
chrisahardie: How were you trying to add an icon?

andrewpmk: I tried icon={LanguageMinor} and the icon doesn't show up - was hoping we could put an icon beside Language so it's obvious even if you don't speak English what it means

andrewpmk: translating the string "Language" does work though

src/i18n.ts
```diff
@@ -1,40 +1,89 @@
...
-          if (lng?.substring(0, 2) === "fr") {
-            return date_fnz_tz_format(value, fmt, { locale: frCA });
-          }
-          return date_fnz_tz_format(value, fmt, { locale: enCA });
-        }
-        return value;
-      },
-    },
+  .init(i18nconfig, function callback() {
+    // We use zh-CN for Simplified Chinese and zh-TW for Traditional Chinese
```
chrisahardie, line 78: Wow, i'm glad I've only ever had to support localizing English and French

andrewpmk: I only need this for Chinese because there are two versions Simplified and Traditional. We only have Latin script for Serbian if we get Cyrillic script we'll need to do this. Same with Punjabi which has two scripts which we don't have yet. Most other languages we don't need this. e.g. we don't implement separate translations for en-CA and en-US only difference is minor things like colour vs color.

src/AppFrame.test.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { render, screen } from "./testUtils";
+import { AppFrame } from "./AppFrame";
+import userEvent from "@testing-library/user-event";
+
+describe("AppFrame", () => {
+  test("Should switch to French", async () => {
+    render(<AppFrame />);
+    const francais = await screen.findByText("Français");
+    userEvent.click(francais);
+    await screen.findByText("Langue"); // Look for lowercase here, CSS is used to change text to uppercase
```
chrisahardie, line 11: No action required, but there is a Chrome extension called "Testing Playground" that lets you inspect an element in devtools and supplies the best possible query to find an element with testing-library. In this case it would have suggested `/langue/i` - you can use regex to query elements and ignore case.

andrewpmk: I only bothered to test French here and not any other languages here. Just wanted to write a basic unit test here to test that the language toggle works. New to this and testing-library is rather slow

src/AppFrame.tsx
```diff
@@ -1,28 +1,130 @@
...
+    const selected =
+      i18next.language === language.langCode ||
+      i18next.language?.substring(0, 2) === language.langCode;
+    return {
+      url: "/",
+      label: language.label,
+      onClick: () => {
+        i18next
+          .changeLanguage(language.langCode)
+          .then(() => {
```
chrisahardie, line 109: These are gratuitous. You don't need the `then`, but you may need to the `catch` IIRC because this would be a floating promise otherwise. A `console.error` statement could be added until we add logging - or we could add actual logging with winston or something - but this is fine for the time being.

andrewpmk: I did this solely to stop eslint from complaining.

chrisahardie (approved):

src/AppFrame.test.tsx
chrisahardie: Yes, and this was a good test. Just wanted to point you in the direction of a helpful tool. All libs that mount  components to a virtual DOM are unfortunately somewhat slow.

src/AppFrame.tsx
chrisahardie: Yep - but the purpose of the eslint rule is to prevent a possible error from a async fire and forget call to be swallowed, which makes it difficult to track down.

Issue #84, opened 2021-05-20
Allow for 3 letter postal input on front end


ericherscovich:

chardielxr:
@ericherscovich Should we just constrain to the first three letters given the last three are useless? We would update the regex and probably wouldn't hurt to add the `maxlength=3` attribute to the input element.

MattAlp:
@chardielxr this will also require updating the localization files to let users know to put in the first 3 characters of their postal code

PR #90, opened 2021-05-22
search as component, add to listings page

shapesandobjects:


src/containers/PharmacyList/PharmacyList.tsx
```diff
@@ -190,9 +191,14 @@ export function PharmacyList(props: Props) {
 
   return (
     <>
+      <section aria-label="search" style={{ marginBottom: "2rem" }}>
```
chrisahardie, line 194: Very minor, but there is an inconsistency in that this `section` element isn't in the `Home` component. If it should always be included, it should probably be in the `SearchCard` component.

src/components/SearchCard/SearchCard.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  const invalidPostalCodeMessage = shouldShowInvalidPostal
+    ? t("invalidpostal")
+    : undefined;
+
+  return (
+    <>
+      <Card>
+        <Card.Section>
+          <Form onSubmit={handleSubmit}>
```
chrisahardie, line 29: Hmm, there is a `role="search"` attribute value we could apply to search forms as an assistive cue, but it looks like the `Form` component doesn't support aria props... The `Form` props should probably extend from `IntrinsicAttributes` so as to support attributes assignable to HTML elements.

chrisahardie (dismissed):

ericherscovich (changes requested):
When there are no pharmacies available, the search bar doesn't show up. Look K2T 0E5 which has no availabilities and you'll see this.

Edit: Fixed it.

ericherscovich (approved):

ericherscovich:
Added the 3 digit postal regex to this. Closes #84

ericherscovich (approved):

PR #91, opened 2021-05-22
Appointment as of format updated

ericherscovich:


src/components/PharmacyCard/PharmacyCard.tsx
```diff
@@ -37,6 +37,113 @@ interface PharmacyProps {
...
+    "Aug",
+    "Sept",
+    "Oct",
+    "Nov",
+    "Dec",
+  ];
+  if (secondsSinceCreated < 60) {
+    return "Just Now";
+  }
+  if (secondsSinceCreated > 60 && secondsSinceCreated < 3600) {
```
chrisahardie, line 63: We really should be leveraging `date-fns` when we're working with dates... working with dates directly like this is cumbersome and error prone.

For all these comparisons, `isAfter()`, `isBefore`, `addSeconds`, `addMinutes` etc would result in a simpler implementation.

PR #95, opened 2021-05-25
Admin portal login and location details

NeilOkolowsky:


src/components/Admin/AddPopUpClinic/AdminPopUpClinic.tsx
```diff
@@ -0,0 +1,152 @@
...
+export interface IPopUpClinic {
+  name: string;
+  date: string;
+  street: string;
+  city: string;
+  province: string;
+  postalCode: string;
+  phone: string;
+  websiteUrl: string;
+  [key: string]: string;
```
chrisahardie, line 16: I don't think you need to index this object by arbitrary keys, you only want to index by known keys.

src/components/Admin/AddPopUpClinic/AdminPopUpClinic.tsx
```diff
@@ -0,0 +1,152 @@
...
+    date: "",
+    street: "",
+    city: "",
+    province: "",
+    postalCode: "",
+    phone: "",
+    websiteUrl: "",
+  });
+
+  const onFormChanged = (field: string, value: string) => {
```
chrisahardie, line 48: If you change `field: string` to `field: keyof IPopUpClinic` you should be able to remove `[key: string]: string;` in the interface and you'll get type safety on your keys.

src/components/Admin/AddPopUpClinic/AdminPopUpClinic.tsx
```diff
@@ -0,0 +1,152 @@
...
+}
+
+const handleSubmit = () => {
+  return true;
+};
+
+export function AdminPopUpClinic() {
+  const { t } = useTranslation();
+
+  const [popUpClinic, setPopUpClinc] = useState<IPopUpClinic>({
```
chrisahardie, line 26: No action required, but I generally avoid objects for component state (and definitely avoid deeply nested objects due to immutability concerns...) The official guidance suggests splitting out state into individual pieces or pieces of state that change together: https://reactjs.org/docs/hooks-faq.html#should-i-use-one-or-many-state-variables.

src/components/Admin/AddPopUpClinic/AdminPopUpClinic.tsx
```diff
@@ -0,0 +1,152 @@
...
+    date: new Date().toDateString(),
+    street: "",
+    city: "",
+    province: "",
+    postalCode: "",
+    phone: "",
+    websiteUrl: "",
+  });
+
+  const [popUpClinicErrors, setPopUpClincErrors] = useState<IPopUpClinic>({
```
chrisahardie, line 37: All this state is derived from the state in `poUpClinic` - it doesn't need to be stored, I think it just needs to be derived when needed.

src/components/Admin/AddPopUpClinic/AdminPopUpClinic.tsx
chrisahardie, line 48: I think it would make sense to split this function into two - `onFormChanged` should be responsible for updating state, and a second `validate` function which would ensure values are passing business rules. Then your control would look something like:

```
<TextField
  name="postalcode"
  type="text"
  error={validate("postalCode", popUpClinicErrors.postalCode)}
  value={popUpClinic.postalCode}
  label={t("postalcode")}
  onChange={(value) => onFormChanged("postalCode", value)}
/>
```

src/components/Admin/AddPopUpClinic/AdminPopUpClinic.tsx
```diff
@@ -0,0 +1,152 @@
...
+      }
+    }
+
+    if (popUpClinic[field] !== undefined) {
+      setPopUpClinc({ ...popUpClinic, [field]: value });
+    }
+  };
+
+  return (
+    <>
```
chrisahardie, line 76: Very minor, this fragment is not required.

src/components/Admin/AdminLogin/AdminLogin.tsx
```diff
@@ -0,0 +1,57 @@
...
+export function AdminLogin({ setAccessSessionStateToken }: IAdminLoginProps) {
+  const { t } = useTranslation();
+  const [
+    shouldShowInvalidAccessToken,
+    setShouldShowInvalidAccessToken,
+  ] = useState(false);
+  const [accessToken, setAccessToken] = useState("");
+
+  const handleSubmit = () => {
+    const bearerToken = authAccessToken(accessToken);
```
chrisahardie, line 21: It is a convention for functions to have a verb prefix - `authAccessToken` sounds like a variable to me. `getAuthAccessToken` would have a clearer intent.

src/components/Admin/AdminLogin/AdminLogin.tsx
```diff
@@ -0,0 +1,57 @@
...
+      setShouldShowInvalidAccessToken(true);
+    }
+  };
+
+  const invalidAccessToken = shouldShowInvalidAccessToken
+    ? t("invalidaccesstoken")
+    : undefined;
+
+  return (
+    <>
```
chrisahardie, line 35: As an FYI, in case you aren't aware, fragments are only required in order to group sibling elements together - JSX always needs to have a single root element, but `<Card>` could be the root in this case.

src/containers/Admin/AdminContainter.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { AdminPopUpClinic } from "../../components/Admin/AddPopUpClinic/AdminPopUpClinic";
+
+export function AdminContainer() {
+  const [accessToken, setAccessToken] = useState(
+    window.sessionStorage.getItem("bearerToken") || null,
+  );
+
+  if (accessToken && !authAccessToken(accessToken)) {
+    window.sessionStorage.removeItem("bearerToken");
+    setAccessToken(null);
```
chrisahardie, line 13: You may want to add `return <></>;` after this line. `setAccessToken` is an async operation, so there could be a brief render where `<AdminPopUpClinic />` is displayed.

src/containers/Admin/AdminContainter.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { authAccessToken } from "../../utils/auth";
+import { AdminLogin } from "../../components/Admin/AdminLogin/AdminLogin";
+import { AdminPopUpClinic } from "../../components/Admin/AddPopUpClinic/AdminPopUpClinic";
+
+export function AdminContainer() {
+  const [accessToken, setAccessToken] = useState(
+    window.sessionStorage.getItem("bearerToken") || null,
+  );
+
+  if (accessToken && !authAccessToken(accessToken)) {
```
chrisahardie, line 11: Looks like we're hitting the api on every render so long as we have an `accessToken` in state. We may want to check the previous value of `accessToken` and only hit the api if there has been a change. To compare previous and current values, you can use this hook: https://usehooks.com/usePrevious/

src/components/Admin/AddPopUpClinic/AdminPopUpClinic.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  const onFormChanged = (field: string, value: string) => {
+    if (field === "postalCode") {
+      if (!postalCodeIsValid(value)) {
+        setPopUpClincErrors({
+          ...popUpClinic,
+          [field]: t("invalidpostalcode"),
+        });
+      } else {
+        setPopUpClincErrors({ ...popUpClinic, [field]: "" });
```
chrisahardie, line 56: We only have two business rules to validate? Are any of the other fields mandatory?

PR #105, opened 2021-05-31
104 consume endpoint

chrisahardie:


ericherscovich (dismissed):
Approved. Let's wait for backend to also give thumbs up to the query before merging.

PR #120, opened 2021-06-05
118 create rapid appointment

ericherscovich:


src/containers/RapidAppointment/RapidAppointment.tsx
```diff
@@ -1,9 +1,603 @@
...
+      return;
+    }
+
+    let tagsCommaSeperatedString = "";
+
+    if (isCallAheadChecked) {
+      tagsCommaSeperatedString = "Call Ahead";
+    }
+    if (isWalkInChecked) {
+      if (!tagsCommaSeperatedString) {
```
chrisahardie, line 149: Instead of this, it would be better to add all tags to an array. When you're done, call `.join(",")` on the array to concatenate the values separated by commas.

src/containers/RapidAppointment/RapidAppointment.tsx
```diff
@@ -1,9 +1,641 @@
...
+
+    const utcDate = zonedTimeToUtc(
+      startOfDay(new Date()),
+      Intl.DateTimeFormat().resolvedOptions().timeZone,
+    );
+
+    const requestPayload: VaccineAvailabilityExpandedCreateRequest = {
+      active: 1, // boolean indicating if location is active
+      date: format(utcDate, "yyyy-MM-dd'T'HH:mm:ssxxx"),
+      inputType: 1, // represents how availability data was recorded - not used at time of writing
```
ericherscovich, line 221: Let's change this to 2 so we can query for inputType 2 for the removal function

chrisahardie: Updated

src/containers/Login/Login.tsx
```diff
@@ -39,48 +41,98 @@ export function Login() {
...
+          setState({ user });
+          setDidLoginSucceed(true);
+        } else {
+          setIsInvalidCredentials(true);
+        }
+      })
+      .catch((err) => console.error(err));
+  };
+
+  if (!isPharmacist && didLoginSucceed) {
```
chrisahardie, line 61: Don't rely on just a single boolean for redirection - `isPharmacist` is initialized with false, so it may be possible for staff to get redirected with this code block. Add a `isStaff` boolean.

src/containers/Login/Login.tsx
```diff
@@ -39,48 +41,98 @@ export function Login() {
       .catch((err) => console.error(err));
   };
 
-  if (didLoginSucceed) {
-    return <Redirect to="/admin" />;
+  const handlePharmacistSubmit = () => {
+    setIsPharmacist(true);
+    setName("PharmacyUser");
```
chrisahardie, line 46: Why are you setting this and the password down below?

src/containers/Login/Login.tsx
```diff
@@ -39,48 +41,98 @@ export function Login() {
       .catch((err) => console.error(err));
   };
 
-  if (didLoginSucceed) {
-    return <Redirect to="/admin" />;
+  const handlePharmacistSubmit = () => {
+    setIsPharmacist(true);
```
chrisahardie, line 45: You should set this when the promise resolves

src/containers/Login/Login.tsx
```diff
@@ -39,48 +41,98 @@ export function Login() {
...
+    setName("PharmacyUser");
+    setPassword(pharmacyPassword);
```
chrisahardie, line 47: You are setting the pharmacy password to the staff password state.

src/containers/Login/Login.tsx
```diff
@@ -39,48 +41,98 @@ export function Login() {
...
+    post({ name, password })
```
chrisahardie, line 48: This line could be `post({ name: "PharmacyUser", password: pharmacyPassword });` ... i think that would clean up the lines above

src/containers/ExternalKeyInput/ExternalKeyInput.test.tsx
```diff
@@ -0,0 +1,30 @@
+import userEvent from "@testing-library/user-event";
+import React from "react";
+import { render, screen } from "../../testUtils";
+import { ExternalKeyInput } from "./ExternalKeyInput";
+
+describe("ExternalKeyInput", () => {
+  test("Should post correct external key to rapid assessment form", async () => {
+    const { history } = render(<ExternalKeyInput />);
```
chrisahardie, line 8: @johnhok Adding `history` to our custom render was a stroke of genius - really simplifies checking paths and params!!

src/containers/RapidAppointment/RapidAppointment.tsx
```diff
@@ -181,6 +181,98 @@ export function RapidAppointment() {
     return isValid;
   };
 
+  const discordWebhook = () => {
+    const request = new XMLHttpRequest();
```
chrisahardie, line 185: There are few reasons to use `XMLHttpRequest`, `fetch` should be favoured.

src/containers/RapidAppointment/RapidAppointment.tsx
```diff
@@ -181,6 +181,98 @@ export function RapidAppointment() {
...
+    }
+    if (isExpiringDosesChecked) {
+      if (reasoningString === "") {
+        reasoningString += "Expiring Doses";
+      } else {
+        reasoningString += " and Expiring Doses";
+      }
+    }
+
+    let bookingMethodsString = "";
```
chrisahardie, line 204: Have a look at my refactor of `tagsCommaSeparatedString` ... the pattern will save you a bunch of typing.

src/containers/RapidAppointment/RapidAppointment.tsx
```diff
@@ -181,6 +181,98 @@ export function RapidAppointment() {
...
+              inline: true,
+            },
+            {
+              name: "Website",
+              value: `${website}`,
+              inline: true,
+            },
+            {
+              name: "Number Available",
+              value: numAvailable === "" ? "Not Reported" : numAvailable,
```
chrisahardie, line 250: Very minor, but conditions are generally more readable if you check the positive case: `value: numAvailable ? numAvailable : "Not Reported"`

src/containers/RapidAppointment/RapidAppointment.tsx
```diff
@@ -237,6 +329,8 @@ export function RapidAppointment() {
       url: getValidUrl(website),
     };
 
+    discordWebhook();
```
chrisahardie, line 332: This will get invoked in our tests, we need to fix that

chrisahardie (approved):

PR #122, opened 2021-06-09
121 form to remove appointments

chrisahardie:
Creating a draft PR as we will probably want to tweak the verbiage/layout and perhaps add a direct link to it somewhere.


src/containers/RemoveAppointments/RemoveAppointments.tsx
```diff
@@ -0,0 +1,182 @@
...
+  useEffect(() => {
+    if (
+      previousAvailabiltyIdToUpdate !== avaibilityIdToUpdate &&
+      availabilitiesData
+    ) {
+      const availaibilityToUpdate = availabilitiesData.find(
+        (a) => a.id === avaibilityIdToUpdate,
+      )!;
+
+      put({
```
chrisahardie, line 116: I'm not entirely happy with this little dance where I maintain a queue of availability ids to process, and then update them one by one. Unfortunately I don't believe there is a way to make an unknown number of requests in parallel to the same endpoint as `restful-react` binds parameters into the hook on each render, and we cannot create multiple mutators as hooks cannot be placed into loops.

Theoretically I suppose I could create a component that wraps the hook, and then add a component array to this component. Don't know if that would simplify things.

ericherscovich (approved):

PR #123, opened 2021-06-09
Added dose selection, updated PopUpForm

ericherscovich:


src/components/PharmacyCard/PharmacyCard.tsx
```diff
@@ -189,6 +189,12 @@ export function PharmacyCard(props: PharmacyProps) {
         if (tag === "Email") {
           return <Badge status="info">{t("email")}</Badge>;
         }
+        if (tag === "1st Dose") {
```
chrisahardie, line 192: Generally when comparing a variable for many possible values, it is customary to use a `switch` statement.

src/components/PharmacyCard/PharmacyCard.tsx
```diff
@@ -189,6 +189,12 @@ export function PharmacyCard(props: PharmacyProps) {
...
+          return <Badge>{t("firstdose")}</Badge>;
+        }
+        if (tag === "2nd Dose") {
+          return <Badge>{t("seconddose")}</Badge>;
```
chrisahardie, line 196: Every `<Badge>` element should have a unique key attribute: `<Badge key={tag}>` ref: https://reactjs.org/docs/lists-and-keys.html#keys

Instead of returning null, the last element should be `<div key="foobar" />`

chrisahardie: Actually @ericherscovich can you check my update in `chardie/dose-selection`? I don't think we should be using `map` for this, as that projects a collection into a new collection of equal length, which doesn't seem right if we don't have a match for a tag. I updated to a loop instead, can you confirm that works for you?

chrisahardie (approved):

chrisahardie (approved):

pleckey (approved):

PR #129, opened 2021-06-13
Optional appointment type

ericherscovich:


src/containers/RapidAppointment/RapidAppointment.tsx
```diff
@@ -254,6 +247,7 @@ export function RapidAppointment() {
     const discordParams = {
       username: "Pharmacy Updates",
       avatar_url: "https://vaccinehunters.ca/favicon.ico",
+      content: "<@&835240707241148428>",
```
chrisahardie, line 250: This is ok? I don't know anything about Discord, a little googling turned up this for content for a message object:

> contents of the message

Are we publishing this string as a message, or is it a cue or something? Perhaps add a comment so maintainers understand what this is for.

chrisahardie (approved):

src/containers/RapidAppointment/RapidAppointment.tsx
ericherscovich: This make's it so I can do @pharmacy which notifies the pharmacy squad to make a tweet. That long number is the role_id for @pharmacy.
Screen Shot 2021-06-13 at 5 05 42 PM

PR #142, opened 2021-06-16
141 add redirect

chrisahardie:


src/containers/RapidAppointment/RapidAppointment.tsx
```diff
@@ -117,6 +118,16 @@ export function RapidAppointment() {
     }
   }, [locationData, previousData]);
 
+  useEffect(() => {
+    if (isCreateRequestSuccessful) {
```
chrisahardie, line 122: Generally at the end of a flow I would `return <Redirect...` as I *think* it's faster than `useEffect` since `useEffect` enqueues a message in the event loop, but it's not that big of a deal and the way `history` is setup in our tests makes it so convenient to test for path and search params.

ericherscovich (approved):

PR #143, opened 2021-06-16
Webhook fix two

ericherscovich:


src/containers/RapidAppointment/RapidAppointment.tsx
```diff
@@ -287,22 +287,25 @@ export function RapidAppointment() {
             },
             {
               name: "Reasoning",
-              value: reasoningString,
+              value: reasoningString === "" ? "Not Reported" : reasoningString,
```
chrisahardie, line 290: Any chance this could be any other falsy value at runtime? I'm on my phone at the moment, but generally I would write ternary's condition like `!reasoningString ? ...`

chrisahardie (approved):

src/containers/RapidAppointment/RapidAppointment.tsx
ericherscovich: I don't believe so, as it is defined as let `reasoningString = "";` and the only operations that are happening are strings getting appended.

PR #152, opened 2021-06-20
Tweak pharmacy card layout

alistair3149:
- Reduce spacing between badges
- Remove container for banner and reduce spacing
- Format location details into a description list
- Add `address` to localization strings

image

### DIFF
--- a/src/components/PharmacyCard/PharmacyCard.tsx
+++ b/src/components/PharmacyCard/PharmacyCard.tsx
@@ -261,7 +261,7 @@ export function PharmacyCard(props: PharmacyProps) {
     >
       <div data-testid="pharmacy-card">
         <TextContainer>
-          <Stack>{badgeMarkup()}</Stack>
+          <Stack spacing="extraTight">{badgeMarkup()}</Stack>
           <Card.Section fullWidth>{availabilityMarkup()}</Card.Section>
           <Card.Section title={t("details")}>
             <Card.Subsection>{props.address}</Card.Subsection>
--- a/src/components/PharmacyCard/PharmacyCard.tsx
+++ b/src/components/PharmacyCard/PharmacyCard.tsx
@@ -262,7 +262,7 @@ export function PharmacyCard(props: PharmacyProps) {
       <div data-testid="pharmacy-card">
         <TextContainer>
           <Stack spacing="extraTight">{badgeMarkup()}</Stack>
-          <Card.Section fullWidth>{availabilityMarkup()}</Card.Section>
+          {availabilityMarkup()}
           <Card.Section title={t("details")}>
             <Card.Subsection>{props.address}</Card.Subsection>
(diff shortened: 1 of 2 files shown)


src/components/PharmacyCard/PharmacyCard.tsx
```diff
@@ -261,18 +262,22 @@ export function PharmacyCard(props: PharmacyProps) {
...
+          {availabilityMarkup()}
+          <DescriptionList
+            items={[
+              {
+                term: t("address"),
+                description: props.address,
+              },
+              {
+                term: t("phone"),
+                description: props.phone || "-",
```
ericherscovich, line 275: Can we split this description list up so that if `props.phone` is null, which is the case with many pharmacies, it only shows the address? That way it won't say "Phone: -"

alistair3149: Sure I will rewrite the description list into a function then.

alistair3149: I am new to TS. I have been trying for a few hours but the linter keeps flagging error related to unsafe assignment, would you mind to take a look on the following function? I am not sure how to fix it.

```typescript
const detailsMarkup = () => {
    const items: unknown[] = [];
    const keys: string[] = ["address", "phone"];

    keys.forEach((key) => {
      const data = props[key];
      if (data) {
        items.push({
          term: t([key]),
          description: data,
        });
      }
    });
    return items ? <DescriptionList items={items} /> : null;
  };
  ```

Comment (chrisahardie) [reply src/components/PharmacyCard/PharmacyCard.tsx:275]: @alistair3149  `items` is typed to `unknown`, but it should be typed to the shape of the items:

`interface Description {
  term: string;
  description: string;
}

const detailsMarkup = () => {
    const items: Description[] = [];
```

You may actually be able to infer the `items` type from `DescriptionList` which would be better if it is typed:

```
type Descriptions = React.ComponentProps<typeof DescriptionList>["items"];

const detailsMarkup = () => {
    const items: Descriptions = [];
```

alistair3149: Thanks for the suggestion! I am also stuck at trying to retrieve `props.address` and `props.phone` using keys, I got this error instead:
image

chrisahardie: The best way to do this would be to create a `/src/types.ts` file and pop this in it:

`export type ExtractArrayOfKeys<T> = (keyof T)[];`

Then, you should be able to:

```
const keys: ExtractArrayOfKeys<PharmacyProps>= ["address", "phone"];
```
That will provide a cue to TS that those literals are members of the props interface. Otherwise you could add `[propName: string]: any;` as a member to `PharmacyProps`, which will let you index it but by ANY string, which isn't ideal as you lose out on type safety.

PR #159, opened 2021-10-27
Support age tagging

ericherscovich:
Pills on PharmacyCard:
Screen Shot 2021-10-27 at 5 31 15 PM

Checkboxes on RapidAppointments:
Screen Shot 2021-10-27 at 5 32 20 PM

Reporting on Discord:
Screen Shot 2021-10-27 at 5 32 41 PM


src/components/PharmacyCard/PharmacyCard.tsx
```diff
@@ -185,6 +185,20 @@ export function PharmacyCard(props: PharmacyProps) {
       returnTags = [];
       const tags = props.vaccineAvailabilities[availability].tags.split(",");
       tags.forEach((tag) => {
+        if (tag === "5-11 Year Olds") {
```
chrisahardie, line 188: Ideally we would have an enum or union supplied by the API to type this to, string comparisons like this aren't ideal - special characters, typos and casing issues can sometimes lead to bugs.

src/containers/RapidAppointment/RapidAppointment.tsx
```diff
@@ -386,6 +409,14 @@ export function RapidAppointment() {
       tagsCommaSeparatedString.push("3rd Dose");
     }
 
+    if (isChildrenDoses) {
+      tagsCommaSeparatedString.push("5-11 Year Olds");
```
chrisahardie, line 413: Saving this string to a constant would ensure we wouldn't have casing / copy and paste issues.

src/containers/RapidAppointment/RapidAppointment.test.tsx
```diff
@@ -87,6 +87,8 @@ describe("Rapid appointment form", () => {
       /at least one booking method must be checked/i,
     );
     expect(bookingErrors.length).toBe(4);
+    const ageErrors = await screen.findAllByText(/one age group/i);
```
chrisahardie: I'd make the error capture regex more descriptive - "at least one age group must be checked"

Testing Playground makes it super easy to grab a great query.

chrisahardie (approved):

src/components/PharmacyCard/PharmacyCard.tsx
ericherscovich: Yeah I agree, will work with Patrick and Evan to sort out an enum for backend.

PR #161, opened 2021-12-01
160 create push notification form

chrisahardie:


public/service-worker.js
```diff
@@ -0,0 +1,30 @@
...
+}
+
+// Register event listener for the 'push' event.
+self.addEventListener('push', function(event) {
+  // Keep the service worker alive until the notification is created.
+  event.waitUntil(
+    getEndpoint()
+    .then(function(endpoint) {
+      self.registration.showNotification('Find Your Immunization', {
+        body: event.data.text(),
```
chrisahardie: This is where we need to validate - once the BE is firing push notifications we can confirm the structure of the event data so we can serve the notification properly.

public/service-worker.js
```diff
@@ -0,0 +1,30 @@
...
+      });
+    })
+  );
+});
+
+self.addEventListener('notificationclick', function(event) {
+  event.notification.close();
+  event.waitUntil(
+    clients.openWindow("https://appointments.vaccinehunters.ca/search/")
```
chrisahardie: We also need to include the postal code of the pharmacy that is advertising a new availability in the URL.

ericherscovich (approved):

PR #162, opened 2021-12-17
Pharmacist join card

ericherscovich:
Add a card in to direct pharmacists on how to join to self-report.

image


ericherscovich:
Should have used translations and also put this in its own component, but this was requested to be put up ASAP, so it works for now.

chrisahardie (approved):

PR #164, opened 2021-12-20
Make the footer link relative to the searched postal code

ryanwilsonperkin:
When the user has searched for a postal code, we can update the text in the footer to be contextual based on that code. This change introduces that behaviour in a couple of steps:

1. Extract a `Footer` component out of the `AppFrame` and render it within relevant routes (so that it can be passed appropriate values)
2. Based on the postalCode value, render the appropriate link and text
3. Ensure we support the edge case of Nunavut and Northwest Territories which share a postal district

### DIFF
--- a/src/components/Footer/Footer.tsx
+++ b/src/components/Footer/Footer.tsx
@@ -7,81 +7,98 @@ interface LocationData {
   name: string;
 }
 
+/** Given a postal code, fetch the appropriate location in Canada
+ * First letter typically gives the province or territory
+ * Nunavut and Northwest Territories are a notable exception, both starting with 'X'
+ * See: https://en.wikipedia.org/wiki/Postal_codes_in_Canada#Components_of_a_postal_code
+ */
 function getLocation(postalCode: string | undefined): LocationData {
   const rootUrl = "https://vaccinehunters.ca";
-  switch (postalCode?.charAt(0).toUpperCase() ?? "") {
-    case "T":
-      return {
-        url: `${rootUrl}/alberta`,
-        name: "Alberta",
-      };
-    case "V":
-      return {
-        url: `${rootUrl}/britishcolumbia`,
-        name: "British Columbia",
-      };
-    case "R":
-      return {
-        url: `${rootUrl}/manitoba`,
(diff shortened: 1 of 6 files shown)


src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+interface Props {
+  postalCode?: string;
+}
+
+export function Footer({ postalCode }: Props) {
+  const location = getLocation(postalCode);
+  return (
+    <FooterHelp>
+      Get more {location.name} resources{" "}
```
chrisahardie, line 112: @ericherscovich How are we handling localizations now? Should the token to our localization files be added in this PR?\

Otherwise LGTM

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+/** Given a postal code, fetch the appropriate location in Canada
+ * First letter typically gives the province or territory
+ * Nunavut and Northwest Territories are a notable exception, both starting with 'X'
+ * See: https://en.wikipedia.org/wiki/Postal_codes_in_Canada#Components_of_a_postal_code
+ */
+function getLocation(postalCode: string | undefined): LocationData {
+  const rootUrl = "https://vaccinehunters.ca";
+  const postalDistrict = postalCode?.charAt(0).toUpperCase() ?? "";
+  const forwardSortationArea = postalCode?.slice(0, 3).toUpperCase() ?? "";
```
chrisahardie, line 18: TIL that sortation is a word.

src/components/Footer/Footer.tsx
ericherscovich: We did not have a localization for this previously, so nothing more to do in this PR. Now that @andrewpmk is back and we have the large community back, we can probably crowdsource a few more translations and get this translated in a future PR.

ericherscovich (approved):
LGTM, appreciate you writing the tests for it as well and also trapping that edge case. Thank you! Just need to wait on the fix from Patrick on the Azure side and confirm everything is running smoothly in staging :)

PR #168, opened 2021-12-24
Add vaccine info page

ericherscovich:
Add page containing curated resources by our health team on vaccine info.


src/containers/VaccineInfo/VaccineInfo.tsx
```diff
@@ -0,0 +1,125 @@
...
+          }}
+          description="Provides a broad array of vaccine information like information on
+            health and saftey, information for before, during and after your
+            vaccination, and live town halls by experts."
+        >
+          <img
+            alt="This Is Our Shot Logo"
+            width="100%"
+            height="100%"
+            style={{
```
chrisahardie, line 74: I haven't pulled this down, but just be aware these styles can blow an image up, and could make it fuzzy especially on high-res displays.

It's also preferred to put styles in a stylesheet and not inline them.

chrisahardie (approved):

chrisahardie:
I am simulating an external request served over SSL:

image

The data inlined images are served. I've tried a number of CSP policy permutations, I can't get the inlined images to not error out on our deployed instance. Could this be related to how we're hosting? @pleckey I know we can apply CSP policies via an HTTP header by configuring Azure hosting or web.config - do you know if one is being applied?

I'll try applying the most permissive/insecure later today as a final test.

UPDATE - yep, looks like a CSP header is being applied to the page:

image

I was able to adjust the CSP policies in some ways via the CSP meta tag, but not in such a way to allow for data inlined images to render.

andrewpmk:
Looks fine except it's English-only of course.

Trivial issues:
- Change Q&A to Q&amp;A
- Any way that the buttons could open in a new window? Usually you use target="_blank" for that except you're using the MediaCard React component so it looks like you can add external: true to do that

src/containers/VaccineInfo/VaccineInfo.tsx
```diff
@@ -0,0 +1,125 @@
...
+          vaccines.
+        </DisplayText>
+      </Layout.Section>
+      <Layout.Section>
+        <MediaCard
+          portrait
+          title="COVID-19: How vaccines are developed"
+          primaryAction={{
+            content: "Learn more",
+            url: "https://www.youtube.com/watch?v=Y51ZgZCS8J0",
```
andrewpmk, line 27: Links should open in new window use external: true

src/containers/VaccineInfo/VaccineInfo.tsx
```diff
@@ -0,0 +1,125 @@
...
+          />
+        </MediaCard>
+        <MediaCard
+          portrait
+          title="COVID-19 Resource Canada"
+          primaryAction={{
+            content: t("visitwebsite"),
+            url: "https://covid19resources.ca/",
+          }}
+          description="Virtual Q&A sessions and workshops to answer your questions about
```
andrewpmk: Please escape ampersand with &amp;

andrewpmk: I mean `&amp;` because github seems to want to convert it to &

andrewpmk (commented):
I would suggest these changes:

- Change Q&A to Q`&amp;`A
- Change the links to open in an external window

public/index.html
```diff
@@ -1,14 +1,13 @@
...
     <meta charset="utf-8" />
     <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <meta name="theme-color" content="#000000" />
-    <meta
-      name="Find Your Immunization"
-      content="Find Your Immunization by Vaccine Hunters Canada finds you bookable vaccine appointments."
-    />
+    <meta name="Find Your Immunization" content="Find Your Immunization by Vaccine Hunters Canada finds you bookable vaccine appointments." />
+    <meta http-equiv="Content-Security-Policy" content="img-src 'self' 'unsafe-inline' data: https:; script-src *.twimg.com *.twitter.com *.google-analytics.com 'self' 'unsafe-inline'; style-src *.twimg.com *.twitter.com 'self' 'unsafe-inline'; default-src *.twitter.com 'self'; connect-src *.vaccinehunters.ca *.azurewebsites.net 'self';">
```
chrisahardie: Let's get rid of this - I was trying to override policy set by the header. We should have one or the other, and the header is preferred.

public/staticwebapp.config.json
```diff
@@ -1,9 +1,9 @@
...
-  "globalHeaders": {
-    "content-security-policy": "default-src https: 'unsafe-eval' 'unsafe-inline'; object-src 'none'"
-  }
-}
+    "navigationFallback": {
+        "rewrite": "/index.html",
+        "exclude": ["/*.{png,jpg,gif,ico}", "/static/*"]
+    },
+    "globalHeaders": {
+        "content-security-policy": "img-src 'self' 'unsafe-inline' data: https:; script-src *.twimg.com *.twitter.com *.google-analytics.com 'self' 'unsafe-inline'; style-src *.twimg.com *.twitter.com 'self' 'unsafe-inline'; default-src *.twitter.com 'self'; connect-src *.vaccinehunters.ca *.azurewebsites.net 'self';"
```
chrisahardie: Can we get rid of the `*azurewebsites.net` in the `connect-src` whitelist? Or constrain it to a subdomain? I can't recall why i added it, but anyone can make a website in azure and get a domain that satisfies that constraint.

src/containers/VaccineInfo/VaccineInfo.tsx
chrisahardie: Opening a new window can pose accessibility challenges: https://www.w3.org/TR/WCAG20-TECHS/G201.html

chrisahardie (approved):

Issue #173, opened 2021-12-27
Improvement: Selecting a language keeps you on current page


ericherscovich:
Currently, when you are not on the home page, like when you have a postal code selected IE: https://appointments.vaccinehunters.ca/search/k2t, if you select a language it will redirect you to the home page. We should investigate if we can remain on the same page when we select a language.

PR #176, opened 2021-12-27
173 don't redirect to home page on language switch

jennygrahamjones:
By removing `url` from the return of `getLanguageBarItemProps`, the user is not redirected to the home page when they select another language ( as per #173 ). To test, navigate to any page other than the home page and select another language.


ericherscovich (approved):

ericherscovich:
LGTM Thanks!

Issue #174, opened 2021-12-27
Improvement: Make it so users don't have to logout to report another availability.


ericherscovich:
When logged in:
<img width="1354" alt="Screen Shot 2021-12-26 at 10 52 51 PM" src="https://user-images.githubusercontent.com/71022385/147432654-d8097725-7d89-495c-b815-e0d319c62147.png">

You are able to add and remove an appointment, but you can only get to this point once. Also, if you click away from it, like click the home, you have no way to get back to that page, except for logging out and logging back in. We should make it so when logged in there are two buttons, one "Log out", and another "Manage availabilities" menu option which takes you back to the page above (which is the [ExternalKeyInput](https://github.com/Vaccine-Hunters-Canada/VaxFinder-Frontend/blob/main/src/containers/ExternalKeyInput/ExternalKeyInput.tsx)) file.

PR #177, opened 2021-12-27
174 add manage availabilities option when logged in

jennygrahamjones:
Suggested implementation for #174


ericherscovich (approved):
LGTM great work!

PR #178, opened 2021-12-27
Translations of provinces

andrewpmk:
Add translations of provinces e.g. Ontario into French and some other translation updates
Used "context" feature to get this to work so that Ontario is translated as l'Ontario, etc.


ericherscovich (approved):
LGTM thanks!

PR #184, opened 2021-12-28
Support multiple vaccine types from the self-reporting form

### LINKED ISSUE #179: Allow for multiple vaccine types in self-reporting
On [the form](https://github.com/Vaccine-Hunters-Canada/VaxFinder-Frontend/blob/c44c5475d29e6cf51550b0fd58f675e55bc47b39/src/containers/RapidAppointment/RapidAppointment.tsx#L461) where we send vaccine availabilities, we currently only allow for one vaccine type to be selected. This is because our API can only accept a single vaccine type. This means if a pharmacy has Moderna and Pfizer, they have to create two different availabilities for this. We can make a change by sending one POST per vaccine type (as seen in the file linked) for the pharmacy. This means on the form, rather than a dropdown of vaccine type, it will be a checkbox. Then, we send one request per vaccine type they have selected. Since the backend returns all availabilities for a single pharmacy, I believe that our tagging system should pick it up on the frontend with no (or little) change.

ryanwilsonperkin:
Fixes #179 

Changes the vaccine selection from a dropdown to a list of checkboxes to allow for multiple vaccines to be added at once. This allows a pharmacist to add a new set of rapid appointment availability for several different vaccine types without having to submit the form multiple times.

We accomplish it by sending off one request for each vaccine type. If none are selected (or if the "Not Sure" option is selected) then we'll send vaccine type "1". Otherwise each request will use the numeric type of the named vaccine.

**Before**
image

**After**
image


ryanwilsonperkin:
Closing in favour of a branch on the main repo (not fork)

PR #189, opened 2021-12-30
Add more tests

### LINKED ISSUE #63: Update testing
Check for:
* multiple search results returned (available + not available)
* no results return
* error returned
* 1 pharmacy card expanded to show hidden content

andrewpmk:
Add more tests - I'm sure this can be improved. I've created some very basic tests for certain components which only render text like TermsOfService which simply render the component and check that it has rendered and nothing else.
Add scripts test:verbose to run tests with all 
Switch from using substr to substring in postal code
Resolves #63


ericherscovich (approved):
LGTM Thanks

